Issue #90, opened 2014-12-21
GetSummoner() hangs the program


Pawes1234:
In the last update everything worked fine. Now I'm stuck with this:
```c#
var api = RiotApi.GetInstance("my key here");
var summoner = api.GetSummoner(RiotSharp.Region.eune, "ign here");
```
If I'm creating a Console Application, the code runs & works fine.
However, in Windows Forms application, on the GetSummoner line the program just hangs.

lolzitos:
Hi guys, Same problem here, the program get stucked when try to get information from API.

@BenFradet do you need some help to find the problem?

BenFradet:
I pretty much know where the problem is coming from, the thing is I would have to revert the code in master to 783740e49b1127e15e9f8d2d4e03d6ac2f53eebc and I don't have much time to spare right now, unfortunately.

PR #95, opened 2015-01-13
Support for "PRESEASON2015" response in matchhistory

BrendanCosman:
Proposed changes to support the new PRESEASON2015 response for the "season" response in MatchSummary. Not yet documented in the API, but I asked for information here: https://developer.riotgames.com/discussion/riot-games-api/show/uhWkZQWi

Changes have not been compiled or tested locally.


BenFradet:
Nicely done, will merge once I've reverted the issue with #90.

PR #112, opened 2015-02-13
Added missing value to MapType enum

jessehallam:
The non-seasonal (plain) version of Summoner's Rift is inexplicably missing from RiotSharp.MapType enum.


BenFradet:
Thanks for spotting that one, that is indeed weird.

PR #115, opened 2015-02-13
Added Season2015 to StatsEndpoint

diegoaossas:


BenFradet:
Hey,

We only need the new seasons since the map type was added in #112, and according to the documentation, there are only Season3, Season2014 and Season2015.
If you could modify your pr accordingly, that'd be great.

diegoaossas:
Should be ok now

BenFradet:
Yup, looks great. I'll merge and publish a new nuget tonight.

PR #135, opened 2015-06-24
Update RiotSharpExample NuGet packages

JacobSnyder:
RiotSharpExample was using an old verison of the Newtonsoft.Json package
which was causing both to get downloaded. Consolidated to the
Newtonsoft.Json version of the main RiotSharp project.


BenFradet:
Nicely spotted, thanks for your contrib!

Issue #136, opened 2015-07-06
Put API Classes behind Interfaces


Dorgenedge:
This wouldn't change functionality at all, but would allow better use of design principles like Dependency Injection and would allow Mocking for Unit Testing purposes.

PR #137, opened 2015-07-06
Put 3 API Classes behind Interfaces

Dorgenedge:
Basically I just created Interfaces for each of the major API classes. I did this for two reasons:

Firstly, it allows for better use of Dependency Injection (specifically in Unit Testing). I can now Mock these objects to test code that would otherwise depend upon the results of API calls.

Secondly, it helps with Inversion of Control a bit.

Any questions or concerns, just lemme know.

Love the project and I'm using it in a personal project!

EDIT: Created an Issue about this earlier, then realized I could do it myself quickly with little effort. If you can associate this with the issue or just remove the issue, go ahead.

### DIFF
--- a/RiotSharp/IRiotApi.cs
+++ b/RiotSharp/IRiotApi.cs
@@ -1,0 +1,80 @@
+﻿using RiotSharp.ChampionEndpoint;
+using RiotSharp.CurrentGameEndpoint;
+using RiotSharp.FeaturedGamesEndpoint;
+using RiotSharp.GameEndpoint;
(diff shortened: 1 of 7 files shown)


BenFradet:
LGTM, I'll try to look into it a bit further today.
I will close the issue when I'll merge (#136).

BenFradet:
Merging, thanks for your contribution!

Dorgenedge:
When is your next scheduled NuGet push? If you don't mind me asking.

BenFradet:
I'll try to do one tonight.

Issue #119, opened 2015-02-27
Convert the library to a portable class library


BenFradet:
As per #117, it'd be cool to make the library compatible with windows phone apps.

PR #165, opened 2015-12-30
[RS-119] Convert the library to a portable class library

BenFradet:
cf #119 

The aim is to make RiotSharp compatible with Windows Phone app development as well as Windows Store app development.

  - [ ] ~~replace monitor by lock~~ won't do as `TryEnter` is useful
  - [x] unit tests for the new cache


BenFradet:
1 build fails because of parallel builds.

Issue #203, opened 2016-03-01
HandleWebException is now useless


BenFradet:
Since we're now using Microsoft.Net.Http (cf #194), [HandleWebException](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Requester.cs#L158) is now useless.

We'll have to replace it and reinstate the ignored tests:
  - [StaticRiotApiExceptionTest](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharpTest/StaticRiotApiExceptionTest.cs)
  - [RiotApiExceptionTest](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharpTest/RiotApiExceptionTest.cs)

PR #211, opened 2016-03-19
Remove exception handlers from Requester

Ostrea:
As said in #203 I have removed try/catch and replaced it with error handling based on response status code. After that it has started to pass tests for expected exception.


BenFradet:
Great, I'll have a look later today or tomorrow!

BenFradet:
LGTM, thanks a lot for your contrib!

Ostrea:
Glad to hear it

Issue #185, opened 2016-01-13
Lobby events' event type are not of represented by the right enum


BenFradet:
The [TournamentLobbyEvent's EventType](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/TournamentEndpoint/TournamentLobbyEvent.cs#L19) could be an enum instead of a simple string.

There is a list of event types in [Riot's documentation](https://developer.riotgames.com/docs/tournaments-api) (towards the bottom of the page).
I've asked Riot if it is exhaustive, will update this issue when I get a response.

In the meantime, we can create a new enum based on this list.

PR #213, opened 2016-03-22
Change lobby events' event type

Ostrea:
Change its type from string to enum as said in #185. I would like to test it, but I don't have tournament api key, so I would be glad if you can test it on your end.

### DIFF
--- a/RiotSharp/RiotSharp.csproj
+++ b/RiotSharp/RiotSharp.csproj
@@ -246,6 +246,7 @@
     <Compile Include="TeamEndpoint\Team.cs" />
     <Compile Include="TeamEndpoint\TeamMemberInfo.cs" />
     <Compile Include="TeamEndpoint\TeamStatDetail.cs" />
+    <Compile Include="TournamentEndpoint\LobbyEvent.cs" />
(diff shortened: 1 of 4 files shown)


BenFradet:
We usually put the enums in their own file like for [TournamentMapType](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/TournamentEndpoint/Converters/TournamentMapType.cs).

Ostrea:
Will move it to different file. Should I put JsonConverter on enum or on property?

Ostrea:
Moved to its own file and placed JsonConverter on the type

BenFradet:
LGTM, could you add some documentation like in the previously linked TournamentMapType?

Ostrea:
Will do it tomorrow

BenFradet:
Sure, no problem.

Ostrea:
I have added documentation.

BenFradet:
LGTM

However, I'd prefer the name of the enum to be `TournamentEventType` to avoid confusions.
Also, the enum should be in the `Converters` folder like the other enums.

Ostrea:
Done

BenFradet:
LGTM merging, thanks.

Issue #162, opened 2015-12-28
Reduce the number of warnings


BenFradet:
Reduce the number of warnings outputted by the ci.
For example: https://ci.appveyor.com/project/BenFradet/riotsharp/build/2.3.3.13

Ostrea:
I am not sure that it helped, according to [this](https://ci.appveyor.com/project/BenFradet/riotsharp/branch/develop) there are still 1591 warnings

BenFradet:
The messages are [here](https://ci.appveyor.com/project/BenFradet/riotsharp/branch/develop/messages)

Ostrea:
Yeah and I still have all warnings related to missing docs

BenFradet:
Yeah [this line](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/RiotSharp.csproj#L29) should have been `<NoWarn>CS1591</NoWarn>`.

I didn't review your changes carefully enough sorry.

Ostrea:
Are you sure about it? Stackoverflow says it should be just 1591

BenFradet:
Welp, I'll try out both tomorrow.

Ostrea:
Also please try /p:nowarn=1591 as paramater for msbuild

BenFradet:
I thought about it and I think we should limit the removal of warnings only regarding the following classes:

  - [IRiotApi](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/IRiotApi.cs)
  - [IStaticRiotApi](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/IStaticRiotApi.cs)
  - [IStatusRiotApi](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/IStatusRiotApi.cs)
  - [ITournamentRiotApi](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/ITournamentRiotApi.cs)

with:

```c#
#pragma warning disable 1591
// the interfaces
#pragma warning restore 1591
```

because the others are legitimate warnings.

Ostrea:
Will do it. But will it stop CI from showing them?

BenFradet:
It should, yes.

Ostrea:
I have done it and made new pull request #216

BenFradet:
Great, I'll have a look later today.

PR #214, opened 2016-03-22
Reduce the number of warnings

Ostrea:
Suppress warning 1591 to remove warning clutter described in #162. Also fix a couple of related warnings.


BenFradet:
Nicely done!

PR #216, opened 2016-03-24
Suppress warning 1591 in certain files

Ostrea:
Only suppress warning 1591 in certain files and not in the whole project. Files in which to suppress are discussed in #162


BenFradet:
Great, merging!

Issue #215, opened 2016-03-23
Replace all Converters folder by Enums


BenFradet:
In the different endpoints, there is a `Converters` folder like for [the tournament api](https://github.com/Ostrea/RiotSharp/tree/ChangeLobbyEventType/RiotSharp/TournamentEndpoint/Converters).

Since those folders contain both enums and associated json converters, I think it'd be better if they'd be named `Enums`.

Ostrea:
Maybe split them to two different folders?

BenFradet:
Yup, it might indeed be better to split them into two different folders.

Ostrea:
So one folder `Converters` with converters and other one `Enums` with enums?

BenFradet:
What would be even better, I think, would be to put the converter in the same file as the enum and have only one `Enums` folder.

What do you think?

Ostrea:
They are quite big, I think it will clutter enums definitions. What about placing `Converters` folder inside `Enums` folder?

BenFradet:
Yup, that seems like the best thing to do.

PR #219, opened 2016-03-28
Replace all Converters folder by Enums

Ostrea:
As discussed in #215 move enums definition to folder named `Enums` and to namespace `Enums`. Also move converters to folder named `Converters` which is inside `Enums` and to namespace `Enums.Converters`


BenFradet:
LGTM, merging.

Thanks a lot.

PR #233, opened 2016-06-10
Rename none to basic on all static data enums

### LINKED ISSUE #231: Rename none to basic in the static api
The `none` enum values for the static API in [DataRequested](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/StaticDataEndpoint/DataRequested.cs) should be renamed to `default` as this is a more fitting name.

shrayasr:
`basic` makes more sense as compared to `none`. None gives a sense of
"nothing" but what this actually does is that it returns a basic set of
keys in the response, so `basic` makes more sense.

Fixes #231


BenFradet:
LGTM, merging. Thanks!

PR #237, opened 2016-06-11
Fix Queue's toCustomString 

MSAlih1:
when i use "JsonConvert.DeserializeObject<RiotSharp.MatchEndpoint.MatchList>(jsoni)", i got null reference error. i hope now working good.


BenFradet:
Isn't it because a new queue type was added?
Which endpoint are you using?

MSAlih1:
RiotSharp.MatchEndpoint.MatchList
but i see this fix not working correctly :/  
im trying

BenFradet:
Thanks for spotting that, merging!

Issue #163, opened 2015-12-30
Unit tests for the Requester


BenFradet:
Unit tests are required for the [Requester class](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Requester.cs).

BenFradet:
Additionally, mark the Requester class as internal explicitly.

shrayasr:
I'm new to writing **proper** unit tests. Could you guide me through what tests would be required here? I was thinking something like this: 

- Make a request to fetch a champion. Check that the request returns a json
- Do the same, but in async
- Make a request to a dummy url, check that it throws an exception
- Do the same, but in async

Is this right?

BenFradet:
That's the gist of it, yes.

Basically, for each method in [Requester](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Requester.cs), there should be unit tests.

As an example, if we take [GetResult](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Requester.cs#L50-L65) there should be two tests:

- `GetResult_ShouldSendBackResult_Test()` where you check that a succeeding get request sends back the correct result
- `GetResult_ShouldSendBackStatusCode_Test()` where you check that a failing get request returns the expected status code

shrayasr:
Great. I've something up at #239. Please do give your comments. I'm really bad at writing Unit tests.

lpszBuffer:
I'll take a look at writing the rest of the unit tests for Requester and start this evening. :smile:

lpszBuffer:
Is there a particular reason you haven't been mocking dependencies out of your unit tests?

For example, in RiotApiTest.GetSummoner_ById_Test, we instantiate the RiotApi class, and by proxy instantiate and use a concrete RateLimitedRequester.

So, if this test fails, we don't have any way of knowing whether or not the error we've introduced is in RiotApi.GetSummoner or RateLimitedRequester.CreateGetRequest. In particular this is troubling if we break one of our very low-level dependencies, as the failing tests can bubble up through the entire project leaving us with the impression that "everything broke".

I can continue to write the Requester tests in the fashion of the rest of the project, but it should be with the understanding that these are really integration tests in disguise if we're not mocking out dependent classes.

Alternatively, we could look at introducing a mocking framework like Moq to our tests, but this is not going to be a trivial debt to pay down.

BenFradet:
[...]

PR #239, opened 2016-06-12
Add some unit tests for Requester

shrayasr:
Took an absolutely beginner's whack at it. I'm sure i'm wrong somewhere. Do point out improvement areas. 

Thanks so much!

### DIFF
--- a/RiotSharpTest/RequesterTest.cs
+++ b/RiotSharpTest/RequesterTest.cs
@@ -1,0 +1,71 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using RiotSharp;
+using System;
+using System.Collections.Generic;
+using System.Configuration;
+
+namespace RiotSharpTest
+{
+    [TestClass]
+    public class RequesterTest
+    {
+        private static string apiKey = ConfigurationManager.AppSettings["ApiKey"];
+
+        private static string RootDomain = "global.api.pvp.net";
+        private static string ChampionRootUrl = "/api/lol/static-data/euw/v1.2/champion";
+        private static string ImproperChampionRootUrl = "/api/lol/static-data/euw/v1.2/NOT_A_champion";
+
+        [TestMethod]
+        [TestCategory("Requester")]
+        public void GetResult_ShouldSendBackResult_Test()
+        {
+            var requester = new Requester(apiKey);
+            var json = requester.CreateGetRequest(ChampionRootUrl, RootDomain, new List<string> { $"locale=en_US", "" });
+
+            Assert.IsTrue(json.Length > 0);
+        }
+
+        [TestMethod]
+        [TestCategory("Requester")]
+        public void GetResultAsync_ShouldSendBackResult_Test()
+        {
+            var requester = new Requester(apiKey);
+            var json = requester.CreateGetRequestAsync(ChampionRootUrl, RootDomain, new List<string> { $"locale=en_US", "" });
+
+            Assert.IsTrue(json.Result.Length > 0);
+        }
+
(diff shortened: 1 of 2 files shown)


RiotSharpTest/RequesterTest.cs
```diff
@@ -0,0 +1,71 @@
...
+        private static string ImproperChampionRootUrl = "/api/lol/static-data/euw/v1.2/NOT_A_champion";
+
+        [TestMethod]
+        [TestCategory("Requester")]
+        public void GetResult_ShouldSendBackResult_Test()
+        {
+            var requester = new Requester(apiKey);
+            var json = requester.CreateGetRequest(ChampionRootUrl, RootDomain, new List<string> { $"locale=en_US", "" });
+
+            Assert.IsTrue(json.Length > 0);
```
BenFradet: I think we should compare the content of the json instead of just its length.

RiotSharpTest/RequesterTest.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        [TestMethod]
+        [TestCategory("Requester")]
+        public void GetResultAsync_ShouldSendBackResult_Test()
+        {
+            var requester = new Requester(apiKey);
+            var json = requester.CreateGetRequestAsync(ChampionRootUrl, RootDomain, new List<string> { $"locale=en_US", "" });
```
BenFradet: those lines seems a bit long

BenFradet:
This looks pretty good so far :+1:

shrayasr:
Should we handle cases like `RootDomain` and `RelativeURL` being passed as empty strings?

RiotSharpTest/RequesterTest.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            try
+            {
+                var json = requester.CreateGetRequest(ImproperChampionRootUrl, RootDomain,
+                    new List<string> { $"locale=en_US", "" });
+            }
+            catch (Exception ex)
+            {
+                Assert.IsInstanceOfType(ex, typeof(RiotSharpException));
+            }
```
BenFradet: When expecting an exception, you can add an annotation:

```c#
[ExpectedException(typeof(RiotSharpException))]
```

like in [StaticRiotApiExceptionTest](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharpTest/StaticRiotApiExceptionTest.cs).

BenFradet:
Ideally yes, but I think the tests will become too numerous and it'll be tedious so I don't think it's worth it.

RiotSharpTest/RequesterTest.cs
shrayasr: Ah. I had a feeling that the `IsInstanceOfType` was overkill :)

shrayasr:
I see, makes sense. Allright

shrayasr:
Made changes to use `ExpectedException`

shrayasr:
@BenFradet gentle reminder :bell:

BenFradet:
I thought you were planning on doing the whole suite of tests. I'll merge tonight.

shrayasr:
Ben,

Like I mentioned, I'm sort of a newbie to writing tests, if you can help me
out with what more we need to test, I'd be glad to investigate more and
write them accordingly

Thanks :)

BenFradet:
This is the proper approach for the `GetResult` method. However, as I said in #163, every method needs to be tested.

If you want I can merge this for now and you'll be able to split your work over multiple PRs.

BenFradet:
Closing, since it will be superseded by #253

PR #244, opened 2016-08-22
Add Republic of Korea Slug

shomzy:
Hi,
When i ran the RiotSharpExample project i got an exception on function statusApi.GetShards()
It occurs when de-serializing the List<Shard>, since one of the Shards had a Slug that did not exist. 
In this case it was "kr" for "Republic of Korea".


BenFradet:
Hey,

Thanks for spotting it.

PR #271, opened 2016-11-10
Added missing game subTypes according to actual API reference.

WARpholomey:
The new 2017 season has a flexible queue mode, which is already referenced in the official API. In addition, GameSubType enum lack Siege and Hexakill 6x6 modes.


BenFradet:
Thanks a lot, merging 👍 .

Issue #292, opened 2016-12-18
GetLeagues() throws RiotSharpException("404, Resource not found") when Summoner is unranked.


MingweiSamuel:
When calling GetLeagues on a summoner who is not ranked, RiotSharp throws a RiotSharpException, which makes it hard to tell if the request completely successfully or not. Right now I am checking if the error message starts with "404".

A simple fix would be to store the `HttpStatusCode` as a field of `RiotSharpException` instead of in the message string.

MingweiSamuel:
Also in general 404 errors shouldn't really happen unless they move an API endpoint, so it might be safe to just catch-ignore them

BenFradet:
I agree that the handling of exceptions is a bit clunky.

It's indeed a good idea to store the status code, feel free to open a PR.

MingweiSamuel:
PR won't fully fix the problem. If GetLeagues is called with a bunch of ids and any of the groups (of size MaxNrLeagues = 10) are all unranked, it will cause an exception, and none of the ranks will be returned

PR #294, opened 2016-12-18
RiotSharpException store source HttpStatusCode (new)

MingweiSamuel:
Closed the old pull request (#293) since it was on the wrong branch

#292 (not a full fix)


BenFradet:
This is really good to have anyway imo, merging :+1: .

Issue #291, opened 2016-12-17
Make GetSummonersAsync send requests asnychronously


MingweiSamuel:
Looking at https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/RiotApi.cs#L168

Wouldn't it be possible to make each call to `requester` non-synchronous (remove the `await`) so that all requests get sent out at the same time, then await them all at once before recombining the results?

I tested with a list of almost 4,000 summoner ids. With a direct single call to `GetSummonersAsync`, it takes about 6.3 seconds, but when I group the summoners into groups of 40 and call `GetSummonersAsync` ~100 times all at once, it takes about 2.3 seconds. (Using production API key so the rate limit wasn't an issue.)

Snippet of code I used to test:

    var tasks = SummonerIds.Select((v, i) => new { v, i })
        .GroupBy(x => x.i / 40)
        .Select(x => riotApi.GetSummonersAsync(Region.na, x.Select(y => y.v).ToList()))
        .ToList();
    await Task.WhenAll(tasks);

BenFradet:
Good idea, please open a PR.

MingweiSamuel:
#299

PR #295, opened 2016-12-18
Make grouped async calls asynchronous

MingweiSamuel:
#291

### DIFF
--- a/RiotSharp/RiotApi.cs
+++ b/RiotSharp/RiotApi.cs
@@ -167,17 +167,18 @@ namespace RiotSharp
         /// <returns>A list of summoners.</returns>
         public async Task<List<Summoner>> GetSummonersAsync(Region region, List<long> summonerIds)
         {
-            var list = new List<Summoner>();
-            foreach (var grp in MakeGroups(summonerIds, MaxNrSummoners))
-            {
-                var json = await requester.CreateGetRequestAsync(
-                    string.Format(SummonerRootUrl,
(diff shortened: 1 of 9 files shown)


RiotSharp/RiotApi.cs
```diff
@@ -167,17 +167,18 @@ public List<Summoner> GetSummoners(Region region, List<long> summonerIds)
...
+            var tasks = MakeGroups(summonerIds, MaxNrSummoners).Select(
+                grp => requester.CreateGetRequestAsync(
+                    string.Format(SummonerRootUrl, region.ToString()) +
+                    string.Format(IdUrl, Util.BuildIdsString(grp)), region
+                    ).ContinueWith(
+                        json => JsonConvert.DeserializeObject<Dictionary<long, Summoner>>(json.Result)
+                    )
+                ).ToList();
+
+            await Task.WhenAll(tasks);
```
BenFradet: I'm curious as to what happens if one task were to fail?

MingweiSamuel: The rest of the requests will still be sent, but after all requests complete the `SelectMany(task => task.Result)` would cause the error to be thrown, IIRC

MingweiSamuel:
I accidentally added more commits to the branch for the RateLimiter, not used to Github. Opened #299 which has just the original change (or is it possible to just accept commits one at a time?)

BenFradet:
Ok, closing this in favor of #299

Issue #296, opened 2016-12-19
Rate limiting blocks all regions


MingweiSamuel:
I was looking at the way rate limiting works, and it looks to me that when one region is blocked, it causes other regions to block (there is only one semaphore instance: https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/RateLimitedRequester.cs#L51 )

I wrote a bit of test code below, and the output seems to match.

        [TestMethod]
        [TestCategory("RiotApi")]
        public async Task TestRateLimit()
        {
            var sw = new Stopwatch();
            sw.Start();

            var euwTasks = summonerIds.Take(11).Select(
                s => api.GetSummonerAsync(Region.euw, s));
            var euwTask = Task.WhenAll(euwTasks).ContinueWith(
                res => Console.WriteLine("EUW: " + sw.Elapsed));

            var naTask = api.GetSummonerAsync(Region.na, 69009277).ContinueWith(
                res => Console.WriteLine("NA: " + sw.Elapsed));

            await naTask;
            await euwTask;
        }

.

    NA: 00:00:11.1279019
    EUW: 00:00:11.2118027

Somewhat related issue, the `while(timer) {}` loops are almost certainly sub-optimal unless I am missing something. `HandleRateLimit()` should probably use `Task.Delay(...).Wait()` or `Thread.Sleep(...)`, and there should probably be a `HandleRateLimitAsync()` which uses `await Task.Delay(...)`. I can make a PR if I'm not missing anything.

BenFradet:
Yup, feel free to open a PR.

PR #300, opened 2016-12-20
Use per-region rate limiter

MingweiSamuel:
#296

There are some tests for the RateLimiter which confirm that it is working properly (at least for the 10 second limit). They are pretty janky since it's a time-based component.

RateLimiter also has a `SetRetryAfter(...)` functionality for when requests return a `RetryAfter` header (429) but it is currently not used (would require more changes to the way the RateLimitedRequester is laid out).

### DIFF
--- a/RiotSharpTest/RateLimiterTest.cs
+++ b/RiotSharpTest/RateLimiterTest.cs
@@ -39,10 +39,6 @@ namespace RiotSharpTest
             AssertDelayed(TimeSpan.Zero);
         }
 
-        /// <summary>
-        /// Same as above, but async.
-        /// </summary>
-        /// <returns></returns>
         [TestMethod]
         [TestCategory("RateLimiter"), TestCategory("Async")]
         public async Task SingleRequestAsync()
@@ -82,10 +78,6 @@ namespace RiotSharpTest
             }
         }
 
-        /// <summary>
-        /// Same as above, but async.
-        /// </summary>
-        /// <returns></returns>
         [TestMethod]
         [TestCategory("RateLimiter"), TestCategory("Async")]
         [Timeout(1000 * 10 * 3)]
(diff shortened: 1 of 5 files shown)


RiotSharpTest/RateLimiterTest.cs
```diff
@@ -0,0 +1,312 @@
...
+        public void SingleRequest()
+        {
+            RateLimiter.HandleRateLimit();
+            AssertDelayed(TimeSpan.Zero);
+        }
+
+        /// <summary>
+        /// Same as above, but async.
+        /// </summary>
+        /// <returns></returns>
```
BenFradet: can we remove that?

RiotSharpTest/RateLimiterTest.cs
```diff
@@ -0,0 +1,312 @@
...
+            {
+                RateLimiter.HandleRateLimit();
+                AssertDelayed(delay);
+            }
+        }
+
+        /// <summary>
+        /// Same as above but async.
+        /// </summary>
+        /// <returns></returns>
```
BenFradet: same

RiotSharpTest/RateLimiterTest.cs
```diff
@@ -0,0 +1,312 @@
...
+
+            await Task.WhenAll(tasks);
+            foreach (var task in tasks)
+            {
+                Assert.IsNull(task.Exception);
+            }
+        }
+
+        /// <summary>
+        /// Sends 5 requests. Waits 5 seconds. Sends 10 More requests. Expects first 10 requests to not block, rest to block by 10 seconds.
```
BenFradet: Can you format comment so that they don't exceed 100 chars? It's the guideline we've followed so far.

MingweiSamuel:
Alright, added those changes

BenFradet:
Merging, thanks.

Issue #309, opened 2016-12-31
Event missing monsterSubType


MingweiSamuel:
monsterSubType       string       The monster subtype of the event. Only present if relevant.

Used to distinguish the different types of dragons, afaik.

PR #310, opened 2016-12-31
adding MonsterSubType enum & field

MingweiSamuel:
#309 

Haven't tested yet, will comment when I test.


BenFradet:
Nice, didn't know those had been made available.

MingweiSamuel:
Confirmed that it is working for the 5 dragon subtypes.

BenFradet:
Thanks, merging 👍

Issue #302, opened 2016-12-28
RANKED_PREMADE_3x3 missing in queue enum


proace137:
Got an error when I called the matchlist endpoint when one summoner got "queue":"RANKED_PREMADE_3x3" in his matchlist.

Even in the riot documentation this queue is not included.

This is the call:

https://euw.api.pvp.net/api/lol/euw/v2.2/matchlist/by-summoner/20094017?beginTime=1479440561048&endTime=1479487031100&api_key=###

BenFradet:
Thanks for reporting.

jhwee:
Per https://developer.riotgames.com/docs/game-constants this queue and a few more are deprecated. However, I was able to duplicate the issue -- should we fix?

BenFradet:
Yup, every fix is good to have :+1:

PR #313, opened 2017-01-20
Add RANKED_PREMADE to Queue enums

jhwee:
Addresses RS-302 (#302). However, these values are deprecated per: https://developer.riotgames.com/docs/game-constants


BenFradet:
Nicely done, thanks!

Issue #303, opened 2016-12-28
Region missing in MatchReference


proace137:
Somtimes it can happen that player switched servers / regions (e.g tr to euw).
For those players you have to have the Region property in the MatchReference(matchlist endpoint) to make the right GetMatch call.

BenFradet:
Good idea.

jhwee:
@BenFradet Is this just a matter of adding the Region property into the MatchReference class? 

I guess I don't see how this will get populated with any other value than the region you pass in to the GET request: `/api/lol/{region}/v2.2/matchlist/by-summoner/{summonerId}`

BenFradet:
@jhwee yup :+1: 

> I guess I don't see how this will get populated with any other value than the region you pass in to the GET request: /api/lol/{region}/v2.2/matchlist/by-summoner/{summonerId}

Agreed but we never know

jhwee:
Thanks. I'll do a PR soon. Riot Games API Status shows matchlist-v2.2 down as of now when I ran the unit test. I will try and confirm again later.

PR #314, opened 2017-01-22
[RS-303] Add Region property to MatchReference

jhwee:
Add missing Region property for #303


BenFradet:
Merging, thanks a lot :+1:

PR #323, opened 2017-02-26
Added Assassinate to GameMode.cs

creepercats:

PR #324, opened 2017-02-26
Fixes #320: Added MatchListEndpoint

### LINKED ISSUE #320: Match & Matchlist End Point
Each API endpoint has it's own directory except for MatchEndPoint which contains both match and matchlist endpoint objects.

Should these two be separated? In other words, moving the matchlist stuff into it's own MatchListEndpoint directory.

It may clear up confusion between enums such as Queue and QueueType, and also the matchlist related enums can be moved from Misc into the new endpoint. 

Let me know what you think.

jono-90:


BenFradet:
Do you think we should move Role/RoleConverter and Lane/LaneConverter to Mis?

BenFradet:
Also I think it'd be good to replicate the structure of the other endpoints by having Enums/Converters subfolders even though there is only Queue in this case.

jono-90:
Sounds good, I I'll add these in when I get a chance. 

Are you strict with the namespace naming conventions? I would like to replicate the style conventions for this project as much as possible.

My view is to keep the amount of nested namespaces to a minimum but that means leaving the namespace the same. Do you think the namespace should match the folder structures or happy to leave as is?

Also moving the role/lane to Misc seems like a good idea as they are shared between the two endpoints. Happy to include this.

BenFradet:
I tend to match the folder structure with the namespaces except for the enums which are in a separate folder as they tend to clutter the folder for the endpoint. However, they're still in the same namespace as I don't see the value of them having their own sub-namespace.

BenFradet:
woops

jono-90:
Looking forward to any feedback :)

BenFradet:
LGTM :+1: , could you rebase with master? #323 made changes to GameModeConverter which is why there are conflicts.

jono-90:
Ok this was my first time rebasing, I hope it turned out fine.

BenFradet:
Apparently not, here's what you should do next time:

```bash
git checkout -b my-fix # creation of a specific branch for your fix so you're not in conflict with my develop branch
# make your commits
```

and then when you need to rebase

```bash
git checkout develop
git pull https://github.com/BenFradet/RiotSharp.git develop
git checkout my-fix
git rebase develop
git push -f origin my-fix
```

Your history now seems a bit screwed up.
I think you're better off doing the following:

```bash
git remote add upstream https://github.com/BenFradet/RiotSharp.git # add my remote
git checkout develop2 upstream/develop # develop2 will track this repo's develop
git checkout -b 320-match-list # the branch that will contain your fixes
git cherry-pick 	0135969
git cherry-pick 75e8d54 # This will add the two commits you wrote earlier that contain the fix
git push origin 320-match-list
```

And then open a new PR based on the 320-match-list branch.

Then and only then, you can clean things up by doing:

```bash
git branch -D origin develop # this will remove the develop branch locally
git push origin :develop # this will remove the branch on the remote
git branch -m develop2 develop # we rename develop2 to develop
git push origin develop # now your develop branch tracks mine
```

PR #327, opened 2017-02-28
Add SummonerId to CurrentGame/Participant

tuffff:
Just adding a missing property to the CurrentGameEndpoint/Participant.
Also adding a test for GetCurrentGame, that is only called if SummonerIdWithCurrentGame is set in the App.config.

### DIFF
--- a/RiotSharpTest/RiotApiTest.cs
+++ b/RiotSharpTest/RiotApiTest.cs
@@ -20,7 +20,6 @@ namespace RiotSharpTest
         private static string team2 = ConfigurationManager.AppSettings["Team2Id"];
         private static int gameId = int.Parse(ConfigurationManager.AppSettings["GameId"]);
         private static int championId = int.Parse(ConfigurationManager.AppSettings["ChampionId"]);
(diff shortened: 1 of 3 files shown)


RiotSharpTest/RiotApiTest.cs
```diff
@@ -242,6 +243,30 @@ public void GetChampionAsync_Test()
...
+            if (idWithCurrentgame != 0) {
+                var currentGame = api.GetCurrentGameAsync(Platform.NA1, idWithCurrentgame);
+
+                Assert.IsNotNull(currentGame.Result);
+                Assert.IsTrue(currentGame.Result.Participants.Any(p => p.SummonerId == idWithCurrentgame));
+            }
+        }
+
+        [TestMethod]
+        [TestCategory("RiotApi")]
```
BenFradet: Unfortunately, I don't think it's worth adding unit tests for the current game endpoint as we'll never be able to run them consistently.

tuffff:
Okay, I removed the unit tests from the PR.

BenFradet:
LGTM, merging :+1:

PR #337, opened 2017-04-01
Update MonsterType.cs

tbrickey022:
Added Rift Herald to the enum collection, as it is returned as a monster type for Elite Monster Kills.


tbrickey022:
Done!  Thanks for the fast reply!

BenFradet:
Thanks I'll merge those two commits.

Next time though try to keep your commits in the same PR.

PR #338, opened 2017-04-01
Update MonsterTypeConverter.cs

tbrickey022:
Added support for Rift Herald into Monster Type Converter


BenFradet:
Oops @tbrickey022, this should have been opened against develop and not master, can you create another one?

tbrickey022:
Sorry!  It's been a while since I've used Git.  I'll fix it now.

Issue #331, opened 2017-03-27
Audit I{Static}RiotApi


BenFradet:
`GetMatchList` is missing from [IRiotApi](https://github.com/BenFradet/RiotSharp/blob/master/RiotSharp/IRiotApi.cs), there might be others.

This issue is to check and add the missing methods.

BenFradet:
fixed by b86673de93b8c0bdfc71378b895a3149f0ffc58a

Issue #307, opened 2016-12-29
Change enums to strings for *Type fields


MingweiSamuel:
The enum types, `QueueType`, `GameType`, etc, cause a lot of stability issues because Riot changes the strings returned by the API relatively frequently and without prior notice.

I propose changing the representation of `*Type`s to `string` (instead of enum), and storing the known values as `const`s in static helper classes.

>Hijacking the top post for the table - Ben

|Desired type|Current enum|
|:---:|:---:|
|String|:heavy_check_mark:[GameQueueType](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/CurrentGameEndpoint/Enums/GameQueueType.cs)<br>:heavy_check_mark:[GameSubType](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/GameEndpoint/Enums/GameSubType.cs)<br>:heavy_check_mark:[QueueType](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/MatchEndpoint/Enums/QueueType.cs)<br>[Season](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/MatchEndpoint/Enums/Season.cs)<br>:heavy_check_mark: [Queue](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/MatchListEndPoint/Enums/Queue.cs)<br>:heavy_check_mark: [GameMode](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Misc/GameMode.cs)<br>[Platform](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Misc/Platform.cs)<br>:heavy_check_mark: [Slug](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Misc/Slug.cs)<br>:heavy_check_mark:[ParTypeStatic](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/StaticDataEndpoint/Champion/Enums/ParTypeStatic.cs)<br>~~PlayerStatsSummaryType~~ (Stats endpoint deprecation)|
[...]

PR #360, opened 2017-04-14
Changed GameMode, Queue, Slug to string and added missing GetMatchList declaration

JanOuborny:
Changed types of `GameMode`, `Queue `and `Slug` from enum to string. #307 

Add missing GetMatchList declaration to `IRiotApi`. #331


JanOuborny:
For now it's fine. But thanks for the hint. I will do that for future commits.

PR #366, opened 2017-04-18
Dotnetcore wip

ddurschlag:


BenFradet:
Back to a normal state! :+1:

JanOuborny:
Since we are targeting .NET Standard 1.3 #419 now, a migration to .NET Core isn't needed anymore.

Issue #369, opened 2017-04-21
Tournament API Test not passing in metadata


toothlessG22:
The Tournament API Test does not call the TournamentCreate functions with the proper number of parameters.

Here is an example:
[TournamentRiotApiTest.cs#L245](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharpTest/TournamentRiotApiTest.cs#L245)

Here is the function that is called from the above line:
[TournamentRiotApi.cs#L119](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/TournamentRiotApi.cs#L119)

PR #372, opened 2017-04-21
Added correct number of parameters to Tournament API tests. Fixes Issue #369.

toothlessG22:
Added correct number of parameters to Tournament API tests. Passed an empty list of Longs for missing summonerIds parameter, passed an empty string for the missing metadata parameter.


BenFradet:
Great, thanks a lot 👍 !

PR #380, opened 2017-04-22
Added interfaces for requester. Various namespace changes

JanOuborny:
Rewrote the requester classes:
- Added a `RequesterBase`
- Added requester interfaces: `IRequester`, `IRateLimtedRequester`
- Moved the classes into the `RiotSharp.Http` namespace

Moved api interfaces into the `RiotSharp.Interfaces` namespace.

Changed `Requester` and `RateLimitedRequester` references to `IRequester` and `IRateLimitedRequester`


RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        /// <summary>
+        /// Create a get request and send synchronously it to the server.
+        /// </summary>
+        /// <param name="relativeUrl"></param>
+        /// <param name="region"></param>
+        /// <param name="addedArguments"></param>
+        /// <param name="useHttps"></param>
+        /// <returns>The content of the response.</returns>
+        /// <exception cref="RiotSharpException">Thrown if an Http error occurs. Contains the Http error code and error message.</exception>
```
BenFradet: could we format this so it's easier to read?

RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// The number of allowed calls per 10 seconds
+        /// </summary>
+        int RateLimitPer10S { get; set; }
+        /// <summary>
+        /// The number of allowed calls per 10 minutes
+        /// </summary>
+        int RateLimitPer10M { get; set; }
+
+        /// <summary>
+        /// Create a get request and send synchronously it to the server.
```
BenFradet: `send it synchronously`

RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <param name="region"></param>
+        /// <param name="addedArguments"></param>
+        /// <param name="useHttps"></param>
+        /// <returns>The content of the response.</returns>
+        /// <exception cref="RiotSharpException">Thrown if an Http error occurs. Contains the Http error code and error message.</exception>
+        string CreateGetRequest(string relativeUrl, Region region, List<string> addedArguments = null,
+            bool useHttps = true);
+
+        /// <summary>
+        ///  Create a get request and send asynchronously it to the server.
```
BenFradet: send it async...

RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <param name="region"></param>
+        /// <param name="addedArguments"></param>
+        /// <param name="useHttps"></param>
+        /// <returns>The content of the response.</returns>
+        /// <exception cref="RiotSharpException">Thrown if an Http error occurs. Contains the Http error code and error message.</exception>
+        Task<string> CreateGetRequestAsync(string relativeUrl, Region region,
+            List<string> addedArguments = null, bool useHttps = true);
+
+        /// <summary>
+        /// Create a post request and send synchronously it to the server.
```
BenFradet: send it

It's true for all the others

RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
BenFradet: are we sure we need all of these?

BenFradet: Also true for the other files

RiotSharp/StaticRiotApi.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using Newtonsoft.Json;
+using RiotSharp.Http;
+using RiotSharp.Interfaces;
 using RiotSharp.StaticDataEndpoint;
 using System;
 using System.Collections.Generic;
```
BenFradet: why isn't the StaticRiotApi using `IRequester` instead of `Requester`?

RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
JanOuborny: Ups, sorry that are the standard usings of ths VS class template.

RiotSharp/StaticRiotApi.cs
JanOuborny: Oh that slipped by. Probably didn't throw a build error when I changed the namespaces.

RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
```diff
@@ -0,0 +1,111 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace RiotSharp.Http.Interfaces
+{
+    interface IRateLimitedRequester
```
BenFradet: wouldn't it make sense to have `IRateLimitedRequester` implement `IRequester`?

JanOuborny: This was also my first thought. However, the signatures of the methods don't match. We could do that later on, but that requires a rewrite of the static api endpoint, in which the IRequester is used.

JanOuborny:
Rebased onto develop

RiotSharp/Http/Interfaces/IRateLimitedRequester.cs
```diff
@@ -0,0 +1,111 @@
...
+    public interface IRateLimitedRequester
+    {
+        /// <summary>
+        /// The number of allowed calls per 10 seconds
+        /// </summary>
+        int RateLimitPer10S { get; set; }
+        /// <summary>
+        /// The number of allowed calls per 10 minutes
+        /// </summary>
+        int RateLimitPer10M { get; set; }
```
BenFradet: why is there no `ApiKey` property like in `IRequester`?

JanOuborny: I forgot to add it there, but when I think about it now, I am not sure anymore if we should include it in the interface or not. Technically they are not required for the interface.

BenFradet: either way is fine, I'd just like to keep things consistent :+1:

JanOuborny:
Should be ready to merge now

Issue #378, opened 2017-04-22
Rate Limiter is not based off the return headers from Riot. 429 Errors are not handled.


toothlessG22:

BenFradet:
These two ideas seem pretty good to me.

Regarding the first, does Riot provide useful values in the headers?
iirc, there were pretty useless a while back

BenFradet:
Also this might be two different issues

toothlessG22:
I think the values are pretty useful. They talk about the values [here](https://developer.riotgames.com/rate-limiting.html). There seem to be two ways of doing this. Either don't make the request that will put you over the limit. Or make the request and use the 429 header Retry-After to set a delay until the next call.

The first option is probably the right thing to do, but the second one might be easier to implement.

toothlessG22:
Implementing this, especially for async, is harder than I thought. The problem with checking the X-App-Rate-Limit header is that you need to wait for the response from the server. The point of async is to avoid waiting, so that option does not work. 

Using the Retry-After header might be able to be done by having a static variable in the RateLimiter that stores the latest Retry-Time given by the header. Then on a 429 Error, the requester could update the retry-time and then delay until the try time. However, I don't think that will work because the program will shoot as many requests as it can, and it won't know to stop until the first rate limit exceeding case recieved a response. So, it looks like this is not a great idea to add. 

With the new 429 Error, it's much easier to see that you are going over the rate limit. This allows the user to know that they are being rate limited. The clarity on this error will allow them to be able to find out what is causing it on their end.

MingweiSamuel:
I originally included the [`SetRetryAfter`](https://github.com/BenFradet/RiotSharp/blob/6f10c83e63fc85da7d3f00713f7e2f597164df01/RiotSharp/RateLimiter.cs#L76) method to be used for this situation, but there wasn't easy access to the RateLimiter from the point where http errors were handled so I left it unused. I believe it basically does what is described above (method 2) and it does have the problem that an extra async request will go through as noted in the doc.

If you are running multiple instances of RiotSharp on one API key, it is probably best to manually set the rate limit and divide by the number of concurrent instances, regardless of the 429 issue.

MingweiSamuel:
Another non-fix fix for the extra async request issue is to retry requests (#298)

MingweiSamuel:
Going to take a look at fixing this along with #298

PR #382, opened 2017-04-22
Added 429 Error. Fixes 1/2 of issue #378

toothlessG22:
Adds a 429 Error.


RiotSharp/Requester.cs
```diff
@@ -160,6 +160,8 @@ protected void HandleRequestFailure(HttpStatusCode statusCode)
                     throw new RiotSharpException("404, Resource not found", statusCode);
                 case HttpStatusCode.Forbidden:
                     throw new RiotSharpException("403, Forbidden", statusCode);
+                case (HttpStatusCode)429:
```
BenFradet: I guess it wasn't handled before since it's not defined in `HttpStatusCode`.

PR #384, opened 2017-04-24
Moved status api to v3 (Fixes #344)

### LINKED ISSUE #344: Migrate to lol-status V3
API ref: https://developer.riotgames.com/api-methods/#lol-status-v3

Code ref: https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/StatusRiotApi.cs

toothlessG22:
You now have to use https and an API key for this endpoint. Still not rate limited though. To get the proper URL I had to add the region in front of the root domain. It may be a good idea to add a parameter to the getRequest function that allows for placing the region on the front of the URL like region.api.riotgames.com because that is what Riot has said they are moving all the URLs to.


toothlessG22:
Guess I need a new API key. Oops!

RiotSharpTest/App.config
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <configuration>
     <appSettings>
-        <add key="ApiKey" value="API_KEY"/>
+        <add key="ApiKey" value="28f2fa25-e664-4919-bd2e-2e61caf18c6f"/>
```
BenFradet: can you remove it?

toothlessG22:
Yeah, sorry. I'll do it by the weekend. I've been super busy with school and I have finals this week. Cheers.

BenFradet:
No problem :+1:

RiotSharpExample/App.config
```diff
@@ -14,5 +14,6 @@
         <add key="Team2Id" value="TEAM-0a2ea1b0-f1cf-11e4-bc7e-c81f66ddabda"/>
         <add key="GameId" value="1826202304"/>
         <add key="Region" value="na"/>
+      <add key="Platform" value="na1"/>
```
BenFradet: the indent seems off

RiotSharp/StatusRiotApi.cs
```diff
@@ -14,11 +13,11 @@ public class StatusRiotApi : IStatusRiotApi
     {
         #region Private Fields
         
-        private const string StatusRootUrl = "/shards";
+        private const string StatusRootUrl = "/lol/status/v3/shard-data";
 
-        private const string RegionUrl = "/{0}";
+        private const string RegionSubdomain = "{0}.";
```
BenFradet: Shouldn't this be `PlatformSubdomain`?

RiotSharpTest/App.config
JanOuborny: Don't forget to regenerate it, so it don't get abused.

PR #387, opened 2017-05-01
Added DI constructors to Api endpoints

JanOuborny:
Prepared `Requester `classes and the `Cache `class for DI:

- Set access modifier to `public`
- Improved constructors

Added DI constructors to api endpoints

Fixed `TournamentApiTest `(wrong api key reference)


JanOuborny:
No problem.

PR #395, opened 2017-05-24
Changed links to Documentations!

HurricanKai:
The Links were no longer supported....

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 C# Wrapper for the [Riot Games API](https://developer.riotgames.com/)
 
-Documentation can be found [here]([Documentation](http://benfradet.github.io/RiotSharp/api/index.html)).
+Documentation can be found [here](http://benfradet.github.io/RiotSharp/api/index.html).
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,7 +72,7 @@ foreach (var stat in varusRanked.Stats)
 }
 ```
 
-You can find a list of all the available operations in [IRiotApi](RiotSharp/IRiotApi.cs).
+You can find a list of all the available operations in [TournamentApi in Documentation](http://benfradet.github.io/RiotSharp/api/RiotSharp.RiotApi.html).
```
BenFradet: very minor but could you replace

> TournamentApi in Documentation

by

> TournamentApi in the documentation

?

Same goes for the links below.

HurricanKai:
updated!

README.md
JanOuborny: As far as my English goes, I would just drop `TournamentApi `and make it like this:
`You can find a list of all the available operations in the [documentation] (...).`

This sounds more fluent for me, but I'm not a native speaker.

BenFradet: I think we should still mention the class holding the API being discussed if someone wants to dig into the code without looking at the doc.

JanOuborny: Then I would phrase it like this:
`You can find a list of all the available operations in the [documentation for the TournamentApi] (...).`

PR #396, opened 2017-05-24
Same here as in Dev Branch!

HurricanKai:


BenFradet:
Merging between develop and master only occur during a release

PR #403, opened 2017-05-25
Changed compiler version to C# 6.0

JanOuborny:


appveyor.yml
```diff
@@ -13,7 +13,7 @@ skip_tags: true
 #---------------------------------#
 
 # Operating system (build VM template)
-os: Windows Server 2012
+os: Visual Studio 2015
```
BenFradet: shouldn't this be `image`?

JanOuborny: No, this also works. See for [example](https://github.com/aspnet/Mvc/blob/dev/appveyor.yml)

Issue #371, opened 2017-04-21
Fixing static data values


JanOuborny:
Since the static data endpoint provides a lot of wrong or bad data (which is still persistent in v3). We should provide a function to fix all of this data. Riot is also [recommending this solution](https://discussion.developer.riotgames.com/answers/57/view.html) by themselves.
To start this we should start listing all know issues with the static api.

Known issues:

- Jungle items have no tags (Item ids: 1400, 1401, 1402, 1408, 1409, 1410, 1412, 1413, 1414, 1416, 1418, 1419)
- [Nidalee's skill information is missing](https://discussion.developer.riotgames.com/questions/680/nidalees-skill-information-is-missing.html)
- [Data for Karma's w is wrong](https://discussion.developer.riotgames.com/questions/505/data-for-karmas-w-is-wrong-for-ages.html)

BenFradet:
I don't know about fixing the results ourselves, this doesn't seem like a good solution in the long-term (a lot of maintenance).

JanOuborny:
I know that it comes with a lot of maintenance, therefore I wouldn't integrated completely into the api endpoint. I would rather offer an option for it. So that even if we are behind in updating the code, it doesn't break the whole endpoint.

BenFradet:
I hear what you're saying however I personally think RiotSharp is meant to be "a dumb wrapper" ie people should expect the same things when using RiotSharp or hitting the API directly.

toothlessG22:
I'm not sure if this would be feasible, but what if you allowed the user to add the missing data. Say you had a function that the user could access called overrideChampionStaticData(championId, championDto). You store that championDto so you can return it whenever the user calls for that specific id. I don't know how the caching feature works, but you might be able to leverage that by storing this championDto in the cache.

JanOuborny:
In this case you could make the static api methods `virtual` and let the users do whatever they want with them. However, I like your idea, but I would take a more generic approach. I would
provide a method `RegisterDataProvider(IDataProvider<TEntity> provider)` which registers a provider for the corrected data. This provider gets then called when populating the cache.

JanOuborny:
Closed because of #564

PR #411, opened 2017-05-30
move masteries to v3

mattmast:
Just moving the masteries to V3.

### DIFF
--- a/RiotSharp/RiotApi.cs
+++ b/RiotSharp/RiotApi.cs
@@ -211,33 +211,21 @@ namespace RiotSharp
             return await Task.Factory.StartNew(() => JsonConvert.DeserializeObject<Champion>(json));
         }
  
-        public Dictionary<long, List<MasteryPage>> GetMasteryPages(Region region, List<long> summonerIds)
+        public List<MasteryPage> GetMasteryPages(Region region, long summonerId)
         {
-            var dict = new Dictionary<long, List<MasteryPage>>();
-            foreach (var summonerId in summonerIds)
-            {
-                var json = requester.CreateGetRequest(string.Format(MasteriesUrl, summonerId), region,
-                    usePlatforms: true);
-                var masteries = JsonConvert.DeserializeObject<MasteryPages>(json);
-                dict.Add(summonerId, masteries.Pages);
-            }
-            return dict;
+            var json = requester.CreateGetRequest(string.Format(MasteriesUrl, summonerId), region,
+                usePlatforms: true);
+
+            var masteries = JsonConvert.DeserializeObject<MasteryPages>(json);
+            return masteries.Pages;
         }
  
-        public async Task<Dictionary<long, List<MasteryPage>>> GetMasteryPagesAsync(Region region,
-            List<long> summonerIds)
+        public async Task<List<MasteryPage>> GetMasteryPagesAsync(Region region, long summonerId)
         {
-            var tasks = summonerIds.Select(
-                summonerId => requester.CreateGetRequestAsync(
-                    string.Format(MasteriesUrl, summonerId), region, usePlatforms: true
-                    ).ContinueWith(
-                        json => JsonConvert.DeserializeObject<MasteryPages>(json.Result)
-                    )
-                ).ToList();
+            var json = await requester.CreateGetRequestAsync(string.Format(MasteriesUrl, summonerId), region, 
+                usePlatforms: true);
 
-            await Task.WhenAll(tasks);
-            return tasks.Select(task => task.Result)
-                .ToDictionary(masteries => masteries.SummonerId, masteries => masteries.Pages);
+            return await Task.Factory.StartNew(() => JsonConvert.DeserializeObject<MasteryPages>(json).Pages);
         }
   
         public Dictionary<long, List<RunePage>> GetRunePages(Region region, List<long> summonerIds)
--- a/RiotSharp/RiotApi.cs
+++ b/RiotSharp/RiotApi.cs
@@ -30,7 +30,7 @@ namespace RiotSharp
         private const string SummonerBySummonerIdUrl = "/{0}";
 
         private const string NamesUrl = "/{0}/name";
-        private const string MasteriesUrl = "/{0}/masteries";
+        private const string MasteriesUrl = "/lol/platform/v3/masteries/by-summoner/{0}";
 
         private const string RunesRootUrl = "/api/lol/{0}/v1.4/summoner";
         private const string RunesUrl = "/{0}/runes";
@@ -214,18 +214,12 @@ namespace RiotSharp
         public Dictionary<long, List<MasteryPage>> GetMasteryPages(Region region, List<long> summonerIds)
         {
             var dict = new Dictionary<long, List<MasteryPage>>();
-            foreach (var grp in MakeGroups(summonerIds, MaxNrMasteryPages))
+            foreach (var summonerId in summonerIds)
             {
-                var json = requester.CreateGetRequest(
-                    string.Format(SummonerRootUrl,
-                        region.ToString()) + string.Format(MasteriesUrl, Util.BuildIdsString(grp)),
-                    region);
-                var subDict =
-                    ConstructMasteryDict(JsonConvert.DeserializeObject<Dictionary<string, MasteryPages>>(json));
-                foreach (var child in subDict)
-                {
-                    dict.Add(child.Key, child.Value);
-                }
+                var json = requester.CreateGetRequest(string.Format(MasteriesUrl, summonerId), region,
+                    usePlatforms: true);
+                var masteries = JsonConvert.DeserializeObject<MasteryPages>(json);
+                dict.Add(summonerId, masteries.Pages);
             }
             return dict;
         }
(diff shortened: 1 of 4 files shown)


RiotSharpTest/RiotApiTest.cs
```diff
@@ -137,20 +137,24 @@ public void GetChampionAsync_Test()
         [TestCategory("RiotApi")]
         public void GetMasteryPages_Test()
         {
-            var masteries = api.GetMasteryPages(RiotApiTestBase.summonersRegion, RiotApiTestBase.summonerIds);
+            var fetchCount = Math.Min(10, RiotApiTestBase.summonerIds.Count());
```
BenFradet: Since batching is going away, we can limit ourselves to keeping only 10 summoner ids, what do you think?

RiotSharp/RiotApi.cs
```diff
@@ -212,37 +212,30 @@ public Champion GetChampion(Region region, int championId)
         public Dictionary<long, List<MasteryPage>> GetMasteryPages(Region region, List<long> summonerIds)
```
JanOuborny: Since the api only supports one summoner id per request, I wouldn't differ from Riot's implementation. As partially discussed in #371, RiotSharp is a stupid wrapper and we shouldn't extend the features of the api interface (even if this is only virtually).

RiotSharp/RiotApi.cs
```diff
@@ -212,37 +212,30 @@ public Champion GetChampion(Region region, int championId)
...
-                }
+                var json = requester.CreateGetRequest(string.Format(MasteriesUrl, summonerId), region,
+                    usePlatforms: true);
+                var masteries = JsonConvert.DeserializeObject<MasteryPages>(json);
+                dict.Add(summonerId, masteries.Pages);
             }
             return dict;
         }
  
         public async Task<Dictionary<long, List<MasteryPage>>> GetMasteryPagesAsync(Region region,
```
JanOuborny: same

RiotSharp/RiotApi.cs
mattmast: Sounds good to me.  Wasn't sure if we wanted to break backwards compatibility with RiotSharp, but V3 is breaking a lot of stuff anyway.  I will add the changes.

RiotSharpTest/RiotApiTest.cs
```diff
@@ -135,26 +135,22 @@ public void GetChampionAsync_Test()
...
         {
-            var fetchCount = Math.Min(10, RiotApiTestBase.summonerIds.Count());
-            var masteries = api.GetMasteryPages(RiotApiTestBase.summonersRegion, 
-                RiotApiTestBase.summonerIds.Take(fetchCount).ToList());
+            var pages = api.GetMasteryPages(RiotApiTestBase.summonersRegion, RiotApiTestBase.summoner1Id);
 
-            Assert.IsNotNull(masteries);
-            Assert.AreEqual(fetchCount, masteries.Count);
+            Assert.IsNotNull(pages);
+            Assert.AreEqual(20, pages.Count);
```
JanOuborny: The test will fail if the summoner deletes a mastery page. I would test for a count between 1 and 20.

RiotSharpTest/RiotApiTest.cs
```diff
@@ -135,26 +135,22 @@ public void GetChampionAsync_Test()
 
         [TestMethod]
         [TestCategory("RiotApi")]
-        public void GetMasteryPages_Test()
+        public void GetMasteryPages_Summoner_HasMasteryPages()
```
JanOuborny: Can you change the name to GetSummoner_ExistingSummonerId_ReturnsMasteryPages()? I would be also nice to have a unit test for the case of the usage of an non existing summoner id.

RiotSharpTest/RiotApiTest.cs
```diff
@@ -135,26 +135,22 @@ public void GetChampionAsync_Test()
...
-            Assert.IsNotNull(masteries);
-            Assert.AreEqual(fetchCount, masteries.Count);
+            Assert.IsNotNull(pages);
+            Assert.AreEqual(20, pages.Count);
         }
 
         [TestMethod]
         [TestCategory("RiotApi"), TestCategory("Async")]
-        public void GetMasteryPagesAsync_Test()
+        public void GetMasteryPagesAsync_Summoner_HasMasteryPages()
```
JanOuborny: Change name too

RiotSharpTest/RiotApiTest.cs
```diff
@@ -135,26 +135,22 @@ public void GetChampionAsync_Test()
...
         {
-            var fetchCount = Math.Min(10, RiotApiTestBase.summonerIds.Count());
-            var masteries = api.GetMasteryPagesAsync(RiotApiTestBase.summonersRegion, 
-                RiotApiTestBase.summonerIds.Take(fetchCount).ToList());
+            var pages = api.GetMasteryPagesAsync(RiotApiTestBase.summonersRegion, RiotApiTestBase.summoner1Id);
 
-            Assert.IsNotNull(masteries.Result);
-            Assert.AreEqual(fetchCount, masteries.Result.Count);
+            Assert.IsNotNull(pages);
```
JanOuborny: Test the results not the task.

RiotSharpTest/RiotApiTest.cs
```diff
@@ -135,26 +135,22 @@ public void GetChampionAsync_Test()
...
+            Assert.AreEqual(20, pages.Result.Count);
```
JanOuborny: same

RiotSharpTest/CommonTestBase.cs
```diff
@@ -12,6 +12,7 @@ public class CommonTestBase
         public static Region summoner1and2Region = (Region)Enum.Parse(typeof(Region),
             ConfigurationManager.AppSettings["Summoner1and2Region"]);
 
+        public static long invalidSummonerId = -1;
```
mattmast: I can put this in the app.config if that is preferred.

RiotSharpTest/RiotApiTest.cs
```diff
@@ -135,22 +136,52 @@ public void GetChampionAsync_Test()
...
 
-            Assert.IsNotNull(masteries.Result);
-            Assert.AreEqual(RiotApiTestBase.summonerIds.Distinct().Count(), masteries.Result.Count);
+            Assert.IsNotNull(pages.Result);
+            Assert.IsTrue(pages.Result.Count >= 1 && pages.Result.Count <= 20);
+        }
+
+        [TestMethod]
+        [TestCategory("RiotApi"), TestCategory("Async")]
+        public async Task GetMasteryPagesAsync_InvalidSummonerId_ThrowsResouceNotFound()
```
mattmast: I didn't see any other cases of an async test.  If this needs to be done differently, let me know.

RiotSharpTest/CommonTestBase.cs
JanOuborny: I think that's not needed.

RiotSharpTest/RiotApiTest.cs
BenFradet: should be `public void GetMasteryPagesAsync_InvalidSummonerId_ThrowsResouceNotFound()`

RiotSharpTest/RiotApiTest.cs
```diff
@@ -135,22 +136,52 @@ public void GetChampionAsync_Test()
...
+        {
+            try
+            {
+                await api.GetMasteryPagesAsync(RiotApiTestBase.summonersRegion, RiotApiTestBase.invalidSummonerId);
+                Assert.Fail();
+            }
+            catch (RiotSharpException e)
+            {
+                Assert.AreEqual(e.HttpStatusCode, HttpStatusCode.NotFound);
+            }
```
BenFradet: I think this won't work since you'll most likely get an `AggregateException` wrapping the `RiotSharpException`.

You can have a look at: https://github.com/BenFradet/RiotSharp/blob/c1e799d6113ba49f2736dd58a72c34819e72d443/RiotSharpTest/CommonTestBase.cs#L41-L52 for inspiration

JanOuborny: I updated my `AggregateException` handling code with a recursive method. You can find it [here](https://github.com/BenFradet/RiotSharp/blob/1ead19dd2b242491ffbfc7ea5031e54e7f1a5389/RiotSharpTest/CommonTestBase.cs#L48-L61).

mattmast: I'm thinking about changing `HandleAggregateException` to:
```
void HandleAggregateException(AggregateException exception, Action<RiotSharpException> riotSharpExceptionHandler = null)
```

Which allows for something like:
```
HandleAggregateException(exception, (riotException) =>
{
    Assert.AreEqual(riotException.HttpStatusCode, HttpStatusCode.NotFound);  
});
```
By default it would still use `HandleRiotSharpException` for `EnsureCredibility`.

Thoughts? Good idea? Bad idea?

JanOuborny: Yes, you could do that. You can now rebase your branch onto develop to get that code.

RiotSharpTest/RiotApiTest.cs
```diff
@@ -176,37 +177,67 @@ public void GetChampionAsync_Test()
...
+                Assert.IsNotNull(pages);
+                Assert.IsTrue(pages.Count >= 1 && pages.Count <= 20);
             });
         }
 
-        [Ignore] //Depricated and being refactored in #411
+        [TestMethod]
+        [TestCategory("RiotApi")]
+        public void GetMasteryPages_InvalidSummonerId_ThrowsResouceNotFound()
+        {
```
JanOuborny: Could you implement `EnsureCredibility()` here. I know it won't be need here probably, but the server could still return 500 or 503.

RiotSharpTest/RiotApiTest.cs
```diff
@@ -176,37 +177,67 @@ public void GetChampionAsync_Test()
...
-                Assert.IsNotNull(masteries.Result);
-                Assert.AreEqual(RiotApiTestBase.summonerIds.Distinct().Count(), 
-                    masteries.Result.Count);
+                Assert.IsNotNull(pages.Result);
+                Assert.IsTrue(pages.Result.Count >= 1 && pages.Result.Count <= 20);
             });
         }
 
         [TestMethod]
+        [TestCategory("RiotApi"), TestCategory("Async")]
```
JanOuborny: same here.

PR #415, opened 2017-06-02
Integrated tournament stub api

JanOuborny:


RiotSharp/TournamentRiotApi.cs
```diff
@@ -32,20 +33,23 @@ public class TournamentRiotApi : ITournamentRiotApi
         private static TournamentRiotApi instance;
 
         private readonly IRateLimitedRequester requester;
+        private string tournamentRootUrl;
 
         #endregion
 
-        private TournamentRiotApi(string apiKey, int rateLimitPer10s, int rateLimitPer10m)
+        private TournamentRiotApi(string apiKey, int rateLimitPer10s, int rateLimitPer10m, bool useStub = false)
```
BenFradet: might be useful to update `ITournamentRiotApi` with the associated doc

JanOuborny: But how? The change is constructor related. I could update the related static methods and the DI options method.

BenFradet: my bad, why did we remove the doc for the constructors though?

JanOuborny: I think there was no documentation for the constructor, since it is private and the new DI constructor doesn't really required one until now.

PR #417, opened 2017-06-04
Updated Static Data V3

JanOuborny:
Ref: https://discussion.developer.riotgames.com/articles/1736/updates-to-lol-static-data-v3.html


RiotSharp/StaticDataEndpoint/StatsStatic.cs
```diff
@@ -75,100 +75,100 @@ public class StatsStatic
 
         public double PercentSpellVampMod { get; set; }
 
-        [JsonProperty("rFlatArmorModPerLevel")]
-        public double RFlatArmorModPerLevel { get; set; }
+        [JsonProperty("FlatArmorModPerLevel")]
+        public double FlatArmorModPerLevel { get; set; }
```
BenFradet: The property name seems to match `JsonProperty`, can't we get rid of the `JsonProperty`?

RiotSharp/StaticRiotApi.cs
```diff
@@ -888,7 +895,7 @@ public RealmStatic GetRealm(Region region)
                     string.Format("locale={0}", language.ToString()),
                     summonerSpellData == SummonerSpellData.basic ?
                         string.Empty :
-                        string.Format("spellData={0}", summonerSpellData.ToString())
+                        string.Format("tags={0}", summonerSpellData.ToString())
```
BenFradet: might be worth it to create a variable `tags={0}`, don't you think?

RiotSharp/StaticDataEndpoint/StatsStatic.cs
JanOuborny: Yes, we can 👍

RiotSharp/StaticRiotApi.cs
JanOuborny: Seems logical. I will do that.

Issue #160, opened 2015-12-28
Create a .nuspec


BenFradet:
Create a [.nuspec file](https://docs.nuget.org/create/nuspec-reference) and integrate it with appveyor.

Depends on #158.

JanOuborny:
This is deprecated with the use of the new project format introduced in #419

PR #418, opened 2017-06-04
Created a .nuspec file

JanOuborny:
#160 

Updated assembly info

### DIFF
--- a/RiotSharp/RiotSharp.nuspec
+++ b/RiotSharp/RiotSharp.nuspec
@@ -8,7 +8,7 @@
     <owners>$author$</owners>
     <licenseUrl>https://github.com/BenFradet/RiotSharp</licenseUrl>
     <projectUrl>https://github.com/BenFradet/RiotSharp/blob/develop/LICENSE.md</projectUrl>
-    <iconUrl>https://github.com/BenFradet/RiotSharp/blob/develop/logo.png?raw=true</iconUrl>
+    <iconUrl>https://raw.githubusercontent.com/BenFradet/RiotSharp/develop/logo.png</iconUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <description>$description$</description>
     <releaseNotes>Migrated to V3.</releaseNotes>
--- a/RiotSharp/RiotSharp.nuspec
+++ b/RiotSharp/RiotSharp.nuspec
@@ -11,7 +11,7 @@
     <iconUrl>https://raw.githubusercontent.com/BenFradet/RiotSharp/develop/logo.png</iconUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <description>$description$</description>
-    <releaseNotes>Migrated to V3.</releaseNotes>
+    <releaseNotes>$releaseNotes$</releaseNotes>
     <tags>riot-games riot-games-api league-of-legends</tags>
   </metadata>
 </package>
--- a/RiotSharp/RiotSharp.nuspec
+++ b/RiotSharp/RiotSharp.nuspec
@@ -1,0 +1,18 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>$id$</id>
+    <version>$version$</version>
+    <title>$title$</title>
+    <authors>$author$</authors>
+    <owners>$author$</owners>
+    <licenseUrl>https://github.com/BenFradet/RiotSharp</licenseUrl>
+    <projectUrl>http://PROJECT_URL_HERE_OR_DELETE_THIS_LINE</projectUrl>
+    <iconUrl>http://ICON_URL_HERE_OR_DELETE_THIS_LINE</iconUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>$description$</description>
+    <releaseNotes>Summary of changes made in this release of the package.</releaseNotes>
+    <copyright>Copyright 2017</copyright>
+    <tags>Tag1 Tag2</tags>
+  </metadata>
+</package>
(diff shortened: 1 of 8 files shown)


RiotSharp/RiotSharp.nuspec
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>$id$</id>
+    <version>$version$</version>
+    <title>$title$</title>
+    <authors>$author$</authors>
+    <owners>$author$</owners>
+    <licenseUrl>https://github.com/BenFradet/RiotSharp</licenseUrl>
+    <projectUrl>https://github.com/BenFradet/RiotSharp/blob/develop/LICENSE.md</projectUrl>
```
BenFradet: I think those two are inverted, `licenseUrl` should be `projectUrl` and vice-versa.

RiotSharp/RiotSharp.nuspec
```diff
@@ -0,0 +1,17 @@
...
+    <iconUrl>https://raw.githubusercontent.com/BenFradet/RiotSharp/develop/logo.png</iconUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>$description$</description>
+    <releaseNotes>Migrated to V3.</releaseNotes>
```
BenFradet: mmmh I'm keen on not hardcoding this and letting it be specified when running `nuget pack`, what do you think?

JanOuborny: How can you specify it when running `nuget pack`?

BenFradet: `nuget pack -properties releaseNotes="Migrated to v3"`

JanOuborny: Fixed that in the newest commit

JanOuborny:
I added branch specific AppVeyor configurations. Check it out

appveyor.yml
```diff
@@ -1,96 +1,292 @@
...
 #---------------------------------#
+- 
+  branches:
+    only:
+      - master
 
-# version format
-version: 2.6.0.{build}
+  # version format
+  version: 3.0.0
```
BenFradet: why remove`.{build}`?

appveyor.yml
```diff
@@ -1,96 +1,292 @@
...
+      on_build_success: false
+      on_build_status_changed: false
+  
+  
+  #---------------------------------#
+  #      default configuration      #
+  #---------------------------------#
+- 
+  # version format
+  version: 3.0.0-build{build}
```
BenFradet: isn't there a `build` too many there?

appveyor.yml
```diff
@@ -1,96 +1,292 @@
...
+      to:
+        - benjamin.fradet@gmail.com
+      on_build_failure: true
+      on_build_success: false
+      on_build_status_changed: false
+    - provider: Webhook
+      url: https://webhooks.gitter.im/e/2668b239e8ba3b25f3ff
+      on_build_failure: true
+      on_build_success: false
+      on_build_status_changed: false
```
BenFradet: why duplicate the whole configuration?

appveyor.yml
JanOuborny: No, that's the naming for develop builds. `{builds}` gets parsed into a build number.

appveyor.yml
JanOuborny: Because you don't want it in a master release.

appveyor.yml
JanOuborny: I think this is the only way to have specific configuration for different builds. According to [this](https://www.appveyor.com/docs/branches/) documentation.

appveyor.yml
BenFradet: then why not 3.0.0.x instead of 3.0.0-buildx, the latter doesn't make sense to me

JanOuborny: I'm following the [NuGet docs](https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages#semantic-versioning) and [Semantic Versioning](https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages#semantic-versioning).

BenFradet: let's leave at `3.0.0.{build}`, please, this has been the case since the beginning, let's keep it consistent.

appveyor.yml
BenFradet: the second values will just overwrite the first

JanOuborny: According to the [docs](https://www.appveyor.com/docs/branches/), this shouldn't be the case.

appveyor.yml
JanOuborny: But if you want to push the artifact to NuGet (whhich is the only point of doing this) you will violate the versioning.

appveyor.yml
BenFradet: let's build on tags in master only and keep the yaml to a minimum please.

JanOuborny:
So you want to shut down CI almost completely?

BenFradet:
I'm not sure what you mean?

I think we don't want the same thing, personally I just want something to trigger a build and publish a nuget.

JanOuborny:
@BenFradet Basically it's the same, but I want to trigger the build with a commit to develop and push a NuGet package to a dev feed. However, I still want to build on feature branches, but don't push a package on a commit there. And in master I want to publish a package on a tag. Therefore, I have the different configuration in the `appveyor.yml`.

BenFradet:
Ok, then can  you move branch-specific settings to the branch part and move everything that is shared to the top-level that'll make everything clearer and remove duplicates.

JanOuborny:
I think you can't remove dublicates. Everything must be explicitly configured per configuration, but I cam try it, maybe it works.

JanOuborny:
Closing. No NuGet needed with the new project format.

PR #430, opened 2017-07-03
champions v3

msarsha:
Hey,

Migrated the champions api to v3

Had to change the tests also - tests checked for 10 f2p champions which is no longer correct.

BTW, That's my first PR.

### DIFF
--- a/RiotSharp.Test/RiotApiTest.cs
+++ b/RiotSharp.Test/RiotApiTest.cs
@@ -124,7 +124,6 @@ namespace RiotSharp.Test
         }
 
         [TestMethod]
-        //[Ignore]
         [TestCategory("RiotApi")]
         public void GetChampions_FreeToPlay_Test()
         {
@@ -138,7 +137,6 @@ namespace RiotSharp.Test
         }
 
         [TestMethod]
-        //[Ignore]
         [TestCategory("RiotApi"), TestCategory("Async")]
         public void GetChampionsAsync_FreeToPlay_Test()
         {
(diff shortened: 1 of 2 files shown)


RiotSharp.Test/RiotApiTest.cs
```diff
@@ -133,12 +133,12 @@ public void GetChampions_FreeToPlay_Test()
                 var champions = api.GetChampions(RiotApiTestBase.summoner1and2Region, true);
```
JanOuborny: Can you add the [EnsureCredibility method](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp.Test/RiotApiTest.cs#L25) here and remove the `Ignore` attribute.

RiotSharp.Test/RiotApiTest.cs
```diff
@@ -147,7 +147,7 @@ public void GetChampionsAsync_FreeToPlay_Test()
                 var champions = api.GetChampionsAsync(RiotApiTestBase.summoner1and2Region, true);
```
JanOuborny: Same here.

RiotSharp/RiotApi.cs
```diff
@@ -35,7 +35,7 @@ public class RiotApi : IRiotApi
         private const string RunesRootUrl = "/api/lol/{0}/v1.4/summoner";
         private const string RunesUrl = "/{0}/runes";
 
-        private const string ChampionRootUrl = "/api/lol/{0}/v1.2/champion";
+        private const string ChampionRootUrl = "/lol/platform/v3/champions";
```
JanOuborny: Can you rename this to `ChampionsUrl` so it's consistent with the other v3 urls naming.

msarsha:
Changed the url property name.

The tests are already wrapped with the `EnsureCredibility` method.

JanOuborny:
@msarsha yep, I see it know. Github cut those lines of my view and I overlooked it.

Issue #420, opened 2017-06-08
Refactor Cache to use System.Threading.Timer


JanOuborny:
Code-ref: https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Cache.cs#L267-L271

We are referencing the package System.Reactive.Linq (or Rx-Linq) only for this single method, which I think can be refactor to use `System.Threading.Timer`. This namcespace is part of the `.NetStandard Library` and therefore makes RiotSharp more leightweight.

PR #434, opened 2017-07-05
Refactor Cache to use System.Threading.Timer  #420

ankurMalhotra:


RiotSharp/Cache.cs
```diff
@@ -266,7 +265,12 @@ public int Count()
 
         private void StartObserving<K>(K key, TimeSpan timeSpan)
         {
-            Observable.Timer(timeSpan).Subscribe(x => TryPurgeItem(key));
+            Timer timer = null;
+            timer = new Timer(x =>
+            {
+                TryPurgeItem(key);
+                timer?.Dispose();
+            }, key, timeSpan, TimeSpan.Zero);
```
JanOuborny: Can you replace `TimeSpan.Zero` with `TimeSpan.FromMilliseconds(-1)` to disable periodic signaling. I know that it's technically not needed because of the `timer?.Dispose()`, but that way it is cleaner.

ankurMalhotra:
Cheers for the approval!!  Just wondering why Timespan.Zero wasn't clear. I thought that would have disabled period behavior for the timer as well

JanOuborny:
@ankurMalhotra Actually, I think both works. But -1 makes it more clear,  especially because it's writen in the constructor documentation. The solution with 0 is only briefly stated in the remarks.

Ref: https://msdn.microsoft.com/en-us/library/ah1h85ch(v=vs.110).aspx

PR #439, opened 2017-07-14
Update RateLimiter to use dictionary of rate limits (fixes #437)

### LINKED ISSUE #437: Update rate limit system to use arbitrary time periods
Default development rate limits are now:

     20 requests every 1 seconds
    100 requests every 2 minutes 

Should change the rate limiting system to support an arbitrary set of rate limits/time periods.

MingweiSamuel:
image

Tests pass locally (don't have tournament key). Not sure if CI is updated for the new system so...

### DIFF
--- a/RiotSharp.AspNetCore/RiotSharpOptions.cs
+++ b/RiotSharp.AspNetCore/RiotSharpOptions.cs
@@ -9,16 +9,12 @@
         public RiotSharpOptions()
         {
             RiotApi = new ApiKeyOptions();
-            RiotApi.RateLimitPer10S = 10;
-            RiotApi.RateLimitPer10M = 500;
             TournamentApi = new TournamentApiKeyOptions();
-            TournamentApi.RateLimitPer10S = 10;
-            TournamentApi.RateLimitPer10M = 500;
         }
 
         public bool UseMemoryCache { get; set; }
         public ApiKeyOptions RiotApi { get; set; }
-        public TournamentApiKeyOptions TournamentApi { get; set; }       
+        public TournamentApiKeyOptions TournamentApi { get; set; }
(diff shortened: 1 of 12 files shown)


RiotSharp.Test/appsettings.json
```diff
@@ -1,4 +1,4 @@
 ﻿{
-  "ApiKey": "API_KEY",
+  "ApiKey": "RGAPI-369f27ac-9250-4e30-95e1-6fa1477f0048",
```
BenFradet: !!

MingweiSamuel: Lul

RiotSharp.AspNetCore/RiotSharpOptions.cs
```diff
@@ -9,16 +9,12 @@ public class RiotSharpOptions
         public RiotSharpOptions()
         {
             RiotApi = new ApiKeyOptions();
-            RiotApi.RateLimitPer10S = 10;
-            RiotApi.RateLimitPer10M = 500;
             TournamentApi = new TournamentApiKeyOptions();
```
JanOuborny: Can you add default values here?

MingweiSamuel:
Removed the api key, added default RiotApi rate limits for the options. I don't have any experience with the Tournament API so I don't how those rate limits are. I figure that API isn't meant to be spammed anyway.

JanOuborny:
@MingweiSamuel  The default rate limits for my tournament key are 10 per 10 seconds and 500 per 10 minutes. Can you also add these?

MingweiSamuel:
Updated

BenFradet:
Merging, thanks!

Issue #270, opened 2016-11-07
Use better assertions


BenFradet:
I noticed a few issues:

- `IsTrue(a == b)` instead of `AreEqual(a, b)`
- expected and equal seem to be inverted quite a few times
- `Assert.IsNotNull` before the real assert is useless

PR #440, opened 2017-07-20
#270 use better assertions

EdAllonby:
First off, apologies if I'm not doing this right. Attempting my first pull request :)

Targets the 3 points raised in #270.

I can't verify the 8 tournament tests as I'm still waiting for a tournament API key. However I didn't need to touch these tests so I'm reasonably confident that no harm has been done. All other tests passing as before.


BenFradet:
@MingweiSamuel I have never encountered a null check failing in the 3 years of dev this lib has seen, that's why I wanted to remove them. The problem either occurs before (json deserialization issue) or after (the real check).

MingweiSamuel:
Fair enough.

Issue #347, opened 2017-04-07
Move to the api.riotgames.com domain


BenFradet:
Code ref: https://github.com/BenFradet/RiotSharp/blob/master/RiotSharp/RateLimitedRequester.cs#L25 and everywhere in the file

PR #443, opened 2017-07-25
Replaced all instances of api.pvp.net with api.riotgames.com

oisindoherty:
Closes issue [Move to the api.riotgames.com domain #347](https://github.com/BenFradet/RiotSharp/issues/347)

### DIFF
--- a/RiotSharp/Http/RateLimitedRequester.cs
+++ b/RiotSharp/Http/RateLimitedRequester.cs
@@ -30,7 +30,7 @@ namespace RiotSharp.Http
             if (usePlatforms)
                 rootDomain = GetPlatformDomain(region);
             else
-                rootDomain = region + ".api.pvp.net";
+                rootDomain = region + ".api.riotgames.com";
 
             var request = PrepareRequest(relativeUrl, addedArguments, useHttps, HttpMethod.Get);
 
@@ -48,7 +48,7 @@ namespace RiotSharp.Http
             if (usePlatforms)
                 rootDomain = GetPlatformDomain(region);
             else
-                rootDomain = region + ".api.pvp.net";
+                rootDomain = region + ".api.riotgames.com";
 
             var request = PrepareRequest(relativeUrl, addedArguments, useHttps, HttpMethod.Get);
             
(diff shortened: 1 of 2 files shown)


oisindoherty:
Looked into it and found that the variable wasn't referenced anywhere else in the codebase, so I went ahead and removed it. Unit tests still run fine.

JanOuborny:
I'm afraid this PR doesn't fully meet the requirements. Just changing the main part of the url doesn't lead to valid domains, because the region names are not equal with the platform names.

oisindoherty:
I see, my apologies. I'll replace the references to `region` in the RateLimitedRequester with `GetPlatformDomain(region)`. As far as I can tell, they only show up in the CreatePostRequest and CreatePutRequest functions used solely by the tournament API, which explains why I missed the unit tests on them. Are there any other fixes that I'm missing? 

With regards to "Furthermore, most of the changes will be removed anyway because they will not be hit", would you suggest removing the `usePlatforms` parameter from the [CreateGetRequest function in the RateLimitedRequester?](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Http/RateLimitedRequester.cs#L27-L33) As far as I can tell, all V3 endpoints of the API use the platform names, and all current calls to CreateGetRequest for the V3 endpoints have `usePlatforms` as true.

JanOuborny:
Regardings the `usePlatforms` parameter: It's only purposes was to signal when to use platform domains instead of the region domains. This was neccessary to use v3 and pre-v3 endpoints simultaneously, but now it can be removed. I'm currently working on some migration stuff, so let me do it.

oisindoherty:
Sounds good, I can start working on migrating some endpoints to V3. I've done some preliminary work on CHAMPION-MASTERY-V3, does this conflict with your work?

JanOuborny:
@oisindoherty I'm already done with champion mastery as of now. See #444

oisindoherty:
@JanOuborny Oh gosh, didn't see that. I can start work on implementing the new RUNES-V3 endpoint. It was assigned to @nickwilliamsnewby, but we haven't heard a response in three weeks from him. I'll respond in that issue thread as well.

PR #448, opened 2017-08-01
Raised default sliding expiration time to 1 hour

JanOuborny:
Due to new rate limits to the static endpoint, which are 10 requests per hours, the default sliding expiration time has to be raised to avoid 429 errors.


BenFradet:
merging :+1:

PR #453, opened 2017-08-05
Add basic rate limiting to StaticRiotApi and option to disable cache

### LINKED ISSUE #386: Change enum naming to start with a capital letter
Normaly enums are named with a capital letter. In many enum types this is not the case to match Riot's naming. However, we should change this to match the general coding standards.
To keep the correct value when converting to Riot's naming format just call the enum like this:
`ChampionData.All.ToString().ToLower()`

If there is a case where Riot uses the lower camel case naming, we can build an extension for that.

### LINKED ISSUE #432: Change default request data of static api methods to all
Right now, the default for requested data in `StaticRiotApi` methods is `XData.basic`, e.g. `ChampionData.basic`. I would change that default to `XData.all` to reduce future calls.
Furthermore, I think in the most case you need multiple data types, therefore it's neccessary to use `XData.all` and specififying this can be quiet annoying on multiple calls.

JanOuborny:
- Add basic rate limiting
- Add option to disable caching
- Add option for setting caching time
- Changed default data of static methods from basic to all, fixes #432
- Change enum naming to start with a capital letter, fixes #386

### DIFF
--- a/RiotSharp.AspNetCore/IServiceCollectionExtension.cs
+++ b/RiotSharp.AspNetCore/IServiceCollectionExtension.cs
@@ -2,6 +2,7 @@
 using RiotSharp.Http;
 using RiotSharp.Interfaces;
 using System;
+using System.Collections.Generic;
 
 namespace RiotSharp.AspNetCore
 {
@@ -24,6 +25,9 @@ namespace RiotSharp.AspNetCore
             if (riotSharpOptions.TournamentApi.ApiKey == null && riotSharpOptions.RiotApi.ApiKey == null)
                 throw new ArgumentNullException("No api key provided.", innerException: null);
 
+            if (riotSharpOptions.UseCache && riotSharpOptions.UseMemoryCache)
+                throw new ArgumentException("Only one cache implementation can be enabled at once.");
+
             if (riotSharpOptions.RiotApi.ApiKey != null)
             {
                 var rateLimitedRequester = new RateLimitedRequester(riotSharpOptions.RiotApi.ApiKey,
@@ -32,18 +36,23 @@ namespace RiotSharp.AspNetCore
                     new TournamentRiotApi(rateLimitedRequester));
                 serviceCollection.AddSingleton<IRiotApi>(serviceProvider => new RiotApi(rateLimitedRequester));
 
-                var requester = new Requester(riotSharpOptions.RiotApi.ApiKey);
+                var staticApiRequester = new RateLimitedRequester(riotSharpOptions.RiotApi.ApiKey, new Dictionary<TimeSpan, int>
+                {
+                    { new TimeSpan(1, 0, 0), 10 }
+                });
 
-                if(riotSharpOptions.UseMemoryCache)
+                if (riotSharpOptions.RiotApi.UseMemoryCache)
                 {
                     serviceCollection.AddMemoryCache();
                     serviceCollection.AddSingleton<ICache, MemoryCache>();
                 }
+                else if (riotSharpOptions.RiotApi.UseCache)
+                    serviceCollection.AddSingleton<ICache, Cache>();
                 else
-                    serviceCollection.AddSingleton<ICache, Cache>();            
+                    serviceCollection.AddSingleton<ICache, PassThroughCache>();
 
                 serviceCollection.AddSingleton<IStaticRiotApi>(serviceProvider => 
-                    new StaticRiotApi(requester, serviceProvider.GetRequiredService<ICache>())); 
+                    new StaticRiotApi(staticApiRequester, serviceProvider.GetRequiredService<ICache>(), riotSharpOptions.RiotApi.SlidingExpirationTime)); 
             }
 
             if (riotSharpOptions.TournamentApi.ApiKey != null)
--- a/RiotSharp.AspNetCore/IServiceCollectionExtension.cs
+++ b/RiotSharp.AspNetCore/IServiceCollectionExtension.cs
@@ -25,9 +25,6 @@ namespace RiotSharp.AspNetCore
             if (riotSharpOptions.TournamentApi.ApiKey == null && riotSharpOptions.RiotApi.ApiKey == null)
                 throw new ArgumentNullException("No api key provided.", innerException: null);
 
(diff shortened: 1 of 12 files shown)


RiotSharp.AspNetCore/IServiceCollectionExtension.cs
```diff
@@ -32,18 +36,23 @@ public static IServiceCollection AddRiotSharp(this IServiceCollection serviceCol
...
+                    { new TimeSpan(1, 0, 0), 10 }
+                });
 
-                if(riotSharpOptions.UseMemoryCache)
+                if (riotSharpOptions.UseMemoryCache)
                 {
                     serviceCollection.AddMemoryCache();
                     serviceCollection.AddSingleton<ICache, MemoryCache>();
                 }
+                if (riotSharpOptions.UseCache)
```
BenFradet: why not an else if here since they're mutually exclusive?

RiotSharp.AspNetCore/RiotSharpOptions.cs
```diff
@@ -25,6 +25,13 @@ public RiotSharpOptions()
             };
         }
 
+        /// <summary>
+        /// Enable or disable default caching for static data endpint
+        /// </summary>
+        public bool UseCache { get; set; }
+        /// <summary>
+        /// Enable or disable ASP.NET Core memory cache implemantion for caching
```
BenFradet: implementation

RiotSharp.AspNetCore/RiotSharpOptions.cs
```diff
@@ -25,6 +25,13 @@ public RiotSharpOptions()
             };
         }
 
+        /// <summary>
+        /// Enable or disable default caching for static data endpint
```
BenFradet: the static data endpoint

RiotSharp/StaticRiotApi.cs
```diff
@@ -96,42 +96,55 @@ public class StaticRiotApi : IStaticRiotApi
...
                 apiKey != Requesters.StaticApiRequester.ApiKey)
             {
-                instance = new StaticRiotApi(apiKey);
+                instance = new StaticRiotApi(apiKey, useCache, slidingExpirationTime);
             }
             return instance;
         }
 
-        private StaticRiotApi(string apiKey)
+        private StaticRiotApi(string apiKey, bool useCache = true, TimeSpan slidingExpirationTime = default(TimeSpan))
```
BenFradet: do we think that a TimeSpan of zero is a reasonable default?

RiotSharp/StaticRiotApi.cs
```diff
@@ -96,42 +96,55 @@ public class StaticRiotApi : IStaticRiotApi
...
+            });
             requester = Requesters.StaticApiRequester;
-            cache = new Cache();
+            if (useCache)
+                cache = new Cache();
+            else
+                cache = new PassThroughCache();
+
+            if (slidingExpirationTime != default(TimeSpan))
+                SlidingExpirationTime = slidingExpirationTime;
```
BenFradet: I don't really agree with this, the documentation would confuse people, I prefer to either have a sensible default or provide no default at all.

RiotSharp/StaticRiotApi.cs
```diff
@@ -96,42 +96,55 @@ public class StaticRiotApi : IStaticRiotApi
...
             if (requester == null)
-                 throw new ArgumentNullException(nameof(requester));
+                throw new ArgumentNullException(nameof(requester));
             if (cache == null)
                 throw new ArgumentNullException(nameof(cache));
             this.requester = requester;
             this.cache = cache;
+
+            if (slidingExpirationTime != default(TimeSpan))
+                SlidingExpirationTime = slidingExpirationTime;
```
BenFradet: same

RiotSharp/StaticRiotApi.cs
JanOuborny: I can make it nullable.

RiotSharp/StaticRiotApi.cs
JanOuborny: Doesn't this provide a sensible default?

BenFradet: Why not have `new TimeSpan(1, 0, 0)` as the default value and setting `SlidingExpirationTime` no matter what?

JanOuborny: You can't do that, because `new TimeSpan()` is not a compile-time constant. However I do set it a default in line 89.

BenFradet: but the problem is that the default isn't used...

JanOuborny: It is used inside the methods. The constructor just overrides it if an other value is provided.

BenFradet: Don't you think it's misleading to see a default value of `Timespan(0, 0, 0)` which will in fact never be used?

JanOuborny: Look at the new implementation. I think there is no better way, when you want to use a `TimeSpan` as a parameter.

RiotSharp/StaticRiotApi.cs
```diff
@@ -96,42 +97,59 @@ public class StaticRiotApi : IStaticRiotApi
...
                 apiKey != Requesters.StaticApiRequester.ApiKey)
             {
-                instance = new StaticRiotApi(apiKey);
+                instance = new StaticRiotApi(apiKey, useCache, slidingExpirationTime);
             }
             return instance;
         }
 
-        private StaticRiotApi(string apiKey)
+        private StaticRiotApi(string apiKey, bool useCache = true, TimeSpan? slidingExpirationTime = null)
```
BenFradet: why can't we do:

```c#
private StaticRiotApi(string apiKey, bool useCache = true, TimeSpan slidingExpirationTime = DefaultSlidingExpirationTime)
```
?

If we can't, let's reflect the nullable in `GetInstance` above.

JanOuborny: You can't do that because TimSpan is not a compile-time constant

JanOuborny:
Rebased.

PR #462, opened 2017-08-25
Updated README.md to align with latest build

dgraney:
* Changed how you get a summoner
* Added a deprecated tag to the summoner.GetRankedStats() blurb
* Created a blurb on how to get champion masteries as a means of displaying the interaction between regular and static api. (Not sure if this is necessary, but thought I may as well add ¯\_(ツ)_/¯ ).

Note: for some reason it isn't seeing the change to StatsStatic.cs from before :/

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,25 +55,6 @@ catch (RiotSharpException ex)
 }
 ```
 
-**(DEPRECATED)** To get the stats in ranked for a specific champion for this summoner:
-```c#
-try
-{
-  var varusRanked = summoner.GetStatsRanked(Season.Season3)
-    .Where((s) => s.Name != null && s.Name.Equals("Varus"))
-    .FirstOrDefault();
-}
-catch (RiotSharpException ex)
-{
-  // Handle the exception however you want.
-}
-
-foreach (var stat in varusRanked.Stats)
-{
-  Console.WriteLine(stat.Name + "  " + stat.Value);
-}
-```
-
 You can find a list of all the available operations in [RiotApi in the documentation](http://benfradet.github.io/RiotSharp/api/RiotSharp.RiotApi.html).
 
 ### Tournament API
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ catch (RiotSharpException ex)
 }
 ```
 
-(DEPRECATED) To get the stats in ranked for a specific champion for this summoner:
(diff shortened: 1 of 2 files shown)


RiotSharp/StaticDataEndpoint/StatsStatic.cs
```diff
@@ -8,135 +8,330 @@ namespace RiotSharp.StaticDataEndpoint
     public class StatsStatic
     {
         internal StatsStatic() { }
-
+        
+        /// <summary>
+        /// Flat Armor value for the object.
+        /// </summary>
```
BenFradet: could you remove this commit as it's already been merged?

BenFradet: You should be able to do so with a `git rebase origin/develop` if `origin` is the name of this remote.

README.md
```diff
@@ -47,15 +47,15 @@ To get a summoner:
...
+  var summoner = api.GetSummonerByName(Region.euw, "StopOFlop");
 }
 catch (RiotSharpException ex)
 {
   // Handle the exception however you want.
 }
 ```
 
-To get the stats in ranked for a specific champion for this summoner:
+**(DEPRECATED)** To get the stats in ranked for a specific champion for this summoner:
```
BenFradet: I think we can remove the whole paragraph since people won't be able to do it.

dgraney:
Will make these changes, it may be a bit. A tad busy

RiotSharp/StaticDataEndpoint/StatsStatic.cs
dgraney: You'll have to excuse my lack of knowledge, but I used that command to no avail. the remote is named origin. Any advice?

dgraney: Probably causing more problems than I'm solving :)

dgraney:
Well at the very least, it seems that the commit still exists... however it isn't saying it is a changed file. I'm not sure if that's sufficient. Also deleted that blurb on the ranked stats.

dgraney:
awesome, thanks for the heads up

BenFradet:
Usually it's good practice to have a different branch for each pull request, the flow boils down to:

```bash
git checkout develop
git pull upstream develop # to have a up-to-date copy of develop
git checkout -b my-fix       # creates a new branch dedicated to the fix
...
```

Issue #433, opened 2017-07-05
Move the match API to V3


JanOuborny:
API ref: https://developer.riotgames.com/api-methods/#match-v3
Code ref: https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/RiotApi.cs#L369-L481

dgraney:
Any update on the progress made here?

Might try playing with this, if not.

JanOuborny:
I've done quit a bit. Just use my [match-api](https://github.com/BenFradet/RiotSharp/tree/match-api) branch as the starting point and create a pull request into it, when you are done.

BenFradet:
Last ticket for 3.0.0!

Shidesu:
If @dgraney can't do it, I could finish this soon.

JanOuborny:
@dgraney You can do it if @dgraney doesn't respond until tomorrow. But don't forget to rebase onto my branch.

PR #470, opened 2017-09-25
Match api

Shidesu:
Fixing the issue #433 (based on the match-api branch made by JanOuborny).


BenFradet:
mmh it seems like your git history is dirty, can you checkout from a clean copy of develop and cherry-pick the commits that pertain to the match api?

Shidesu:
Yeah i don't know why the commit is not good.
The copy on my computer is not like that, i'm gonna check.
I think i will reopen a new Pull Request.

PR #477, opened 2017-10-02
Improve string and date Util methods

frederickrogan:


RiotSharp/Misc/Util.cs
```diff
@@ -1,70 +1,50 @@
...
-            string concatenatedIds = string.Empty;
-            for (int i = 0; i < ids.Count - 1; i++)
-            {
-                concatenatedIds += ids[i] + ",";
-            }
-            return concatenatedIds + ids[ids.Count - 1];
+            return string.Join(",", ids);
         }
+
+
```
BenFradet: 2 empty lines instead of one

PR #479, opened 2017-10-03
Null check improvement

maferland:
Found a null check that could be improved. Feel free to decline if you don't like this syntax.


maferland:
Edit: make sure to run the test properly. Most tests passed properly but since I don't know the code base I feel a tad unsafe.

PR #480, opened 2017-10-03
Refactored Requester

frederickrogan:
I have rewritten a version of the standard requester by attempting to separate its responsibilities into new classes. Benefits are improved testability (httpclient can be injected) and I think it's cleaner.

I'd appreciate feedback. If positive I'll continue and attempt to replace the other requesters.

### DIFF
--- a/RiotSharp/Http/Interfaces/IRequestCreator.cs
+++ b/RiotSharp/Http/Interfaces/IRequestCreator.cs
@@ -1,0 +1,11 @@
+﻿using System.Collections.Generic;
+using System.Net.Http;
+using RiotSharp.Misc;
+
+namespace RiotSharp.Http.Interfaces
+{
+    public interface IRequestCreator
+    {
+        HttpRequestMessage CreateGetRequest(Region region, string apiEndpoint, List<string> addedArguments, bool useHttps = true);
+    }
+}
--- a/RiotSharp/Http/ResponseDeserializer.cs
+++ b/RiotSharp/Http/ResponseDeserializer.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using RiotSharp.Http.Interfaces;
+
+namespace RiotSharp.Http
+{
+    public class ResponseDeserializer : IResponseDeserializer
+    {
+        public async Task<T> DeserializeToAsync<T>(HttpResponseMessage message)
+        {
+            var json = await message.Content.ReadAsStringAsync();
+            return JsonConvert.DeserializeObject<T>(json);
+        }
+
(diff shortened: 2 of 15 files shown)


RiotSharp/Http/ResponseDeserializer.cs
```diff
@@ -0,0 +1,22 @@
...
+using RiotSharp.Http.Interfaces;
+
+namespace RiotSharp.Http
+{
+    public class ResponseDeserializer : IResponseDeserializer
+    {
+        public async Task<T> DeserializeToAsync<T>(HttpResponseMessage message)
+        {
+            var json = await message.Content.ReadAsStringAsync();
+            return JsonConvert.DeserializeObject<T>(json);
```
JanOuborny: Couldn't you run this in a task?

RiotSharp/StatusRiotApi.cs
```diff
@@ -6,24 +6,35 @@
 using System.Threading.Tasks;
 using RiotSharp.Misc;
 using System;
+using System.Net.Http;
 
 namespace RiotSharp
 {
-    public class StatusRiotApi : IStatusRiotApi
+    public partial class StatusRiotApi : IStatusRiotApi
```
JanOuborny: Why do you want to split this into two partial classes? I don't think this is really needed.

RiotSharp/Http/Interfaces/IRequestCreator.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using System.Net.Http;
+using RiotSharp.Misc;
+
+namespace RiotSharp.Http.Interfaces
+{
+    public interface IRequestCreator
+    {
+        HttpRequestMessage CreateGetRequest(Region region, string apiEndpoint, List<string> addedArguments, bool useHttps = true);
```
JanOuborny: Here might be needed functions for `PUT` and `POST` for the tournament endpoint.

RiotSharp/Http/ResponseDeserializer.cs
frederickrogan: I did not see the benefit. After a bit of googling I have changed my mind and will put run it in one. Thanks

RiotSharp/Http/Interfaces/IRequestCreator.cs
frederickrogan: You're right. I hadn't written them yet since this only covered the standard requester which is only used by the status api. I'll include it in the branch you suggested.

RiotSharp/StatusRiotApi.cs
JanOuborny: You can split them up into two different files. However, spliting the class up in the same file doesn't really provide a benefit in my eyes. You can achieve the same effect with regions.

JanOuborny:
As far as I can see all changes are in #481. Therefore, I am closing this PR and the discussion can continue in #481.

Issue #459, opened 2017-08-19
Implement deterministic unit testing


JanOuborny:
Mock HttpClient and save api responses as JSON files inside the working tree.

dresdor:
I am starting to take a look at this in my spare time

PR #481, opened 2017-10-04
Requester and RateLimitRequester refactor

frederickrogan:
Updated API classes to use new requester interfaces

PR #483, opened 2017-10-21
Implemented a mock http client for Issue #459

averageprogrammer:
Put the httpclient reference behind a interface to add the ability for
deterministic unit testing of api interactions in the future.

### DIFF
--- a/RiotSharp/Http/Interfaces/IRiotApiClient.cs
+++ b/RiotSharp/Http/Interfaces/IRiotApiClient.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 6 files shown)


JanOuborny:
The problem here is, there is already a pending implementation of a loose coupling of the `HttpClient`:
https://github.com/BenFradet/RiotSharp/pull/481/files#diff-c1fcfa7f3de2007b100a52b0b10ea792R9

We need to merch both of these changes.

averageprogrammer:
Gotcha, didn't realized there was already a pending implementation. Should I do anything on my end to help with the merch of the changes?

BenFradet:
@averageprogrammer yup it'd be nice if you could have a look and review the other PR since you've done something similar :+1:

PR #494, opened 2017-11-22
Add a distributed cache implementation for ASP.NET Core

JanOuborny:
WIP

@BenFradet  What are you thoughts about this?

### DIFF
--- a/RiotSharp.AspNetCore/DistributedCache.cs
+++ b/RiotSharp.AspNetCore/DistributedCache.cs
@@ -1,0 +1,65 @@
+﻿using Microsoft.Extensions.Caching.Distributed;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace RiotSharp.AspNetCore
+{
+    /// <summary>
+    /// Implementation of ICache with AspNetCore's distributed cache
+    /// </summary>
+    public class DistributedCache : ICache
+    {
+        private IDistributedCache distributed;
+        private List<object> usedKeys;
+
+#pragma warning disable CS1591
+        public DistributedCache(IDistributedCache memoryCache)
+        {
+            this.distributed = memoryCache;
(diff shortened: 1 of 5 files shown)


RiotSharp.AspNetCore/IDistributedCacheExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        public static T GetJson<T>(this IDistributedCache distributedCache, string key)
+        {
+            var unserializedValue = distributedCache.GetString(key.ToString());
+            return JsonConvert.DeserializeObject<T>(unserializedValue);
+        }
+    }
+}
+
```
BenFradet: nit: white spaces

RiotSharp.AspNetCore/DistributedCache.cs
```diff
@@ -0,0 +1,54 @@
...
+        private List<object> usedKeys;
+
+#pragma warning disable CS1591
+        public DistributedCache(IDistributedCache memoryCache)
+        {
+            this.distributed = memoryCache;
+            usedKeys = new List<object>();
+        }
+
+        public void Add<K, V>(K key, V value, TimeSpan slidingExpiry) where V : class
```
BenFradet: could we add expliit overrides, for the hybrid cache too?

JanOuborny: What do you mean by explicit overrides?

BenFradet: `public override void Add...`

JanOuborny: I can't override it, because it's not inherited from a virtual method. Furthermore, if override it with the `new` keyword, the method won't fulfill the requirements of the interface.

BenFradet:
good to merge 👍

PR #497, opened 2017-12-08
File caching and caching for match and summoner data

Talha-T:
There are some conflicts though, needs to be fixed and I don't know how.


Talha-T:
I think I need to test this more, I will reopen this when it is ready.

PR #499, opened 2017-12-14
File Caching And Match - Summoner Caching

Talha-T:


RiotSharp/FileCache.cs
```diff
@@ -0,0 +1,80 @@
...
+            return Path.Combine(_directory, key.GetHashCode().ToString());
+        }
+
+        private T Load<T>(string path)
+        {
+            var json = File.ReadAllText(path);
+            return JsonConvert.DeserializeObject<T>(json);
+        }
+
+        private void Store<TK, TV>(TK key, TV value, long ttlMins = 24 * 60 * 7 * 4) // A month
```
BenFradet: Couldn't we type the ttl as a `TimeSpan`?

RiotSharp/FileCache.cs
```diff
@@ -0,0 +1,80 @@
...
+            CacheData<TV> data = new CacheData<TV>(ttlMins, value);
+            File.WriteAllText(GetPath(key.ToString()), JsonConvert.SerializeObject(data));
+        }
+
+        public bool IsExpired<T>(CacheData<T> data)
+        {
+            return data == null || DateTime.Now > data.CreatedAt.AddMinutes(data.TtlMinutes);
+        }
+
+        public void Add<TK, TV>(TK key, TV value, TimeSpan slidingExpiry) where TV : class
```
BenFradet: could we add override to the methods inherited from `ICache`?

RiotSharp/RiotApi.cs
```diff
@@ -63,6 +63,13 @@ public class RiotApi : IRiotApi
         private const string ChampionMasteryBySummonerUrl = "/champion-masteries/by-summoner/{0}/by-champion/{1}";
         private const string ChampionMasteryTotalScoreBySummonerUrl = "/scores/by-summoner/{0}";
 
+        private const string MatchCache = "match-{0}_{1}";
+        private const string SummonerCache = "summoner-{0}_{1}";
+
+        private static readonly TimeSpan SummonerTtl = TimeSpan.FromDays(30);
+        private static readonly TimeSpan MatchTtl = TimeSpan.FromDays(60);
+
+
```
BenFradet: empty line

RiotSharp/RiotApi.cs
```diff
@@ -145,74 +154,110 @@ public RiotApi(IRateLimitedRequester rateLimitedRequester)
...
             {
-                obj.Region = region;
+                var json = requester.CreateGetRequest(
+                    string.Format(SummonerRootUrl + SummonerByAccountIdUrl, accountId), region);
+                var obj = JsonConvert.DeserializeObject<Summoner>(json);
+                if (obj != null)
+                {
+                    obj.Region = region;
+                }
+                cache.Add(string.Format(SummonerCache, region, accountId), obj, SummonerTtl );
```
BenFradet: nit: space before the last parens

RiotSharp/FileCache.cs
Talha-T: We could, but we do not need to store seconds and such, which also enables it to use less memory.

RiotSharp/FileCache.cs
Talha-T: So should I make it virtual void?

RiotSharp/RiotApi.cs
Talha-T: Should I remove line 71?

RiotSharp/RiotApi.cs
Talha-T: Gonna fix this.

Talha-T:
Those should be done now.

RiotSharp/FileCache.cs
BenFradet: public override void since it overrides a method in [ICache](https://github.com/BenFradet/RiotSharp/blob/afb5beabb52e95b29ac1ab103d84a04d08178ae7/RiotSharp/ICache.cs#L19)

RiotSharp/RiotApi.cs
BenFradet: yeah one of those two

RiotSharp/FileCache.cs
Talha-T: *Or disk

RiotSharp/FileCache.cs
Talha-T: I don't think we have to do that.

BenFradet: yeah but it makes it clearer that this method was inherited from ICache

RiotSharp/FileCache.cs
Talha-T: Or you can use `TimeSpan.FromDays(30).TotalMinutes`

PR #508, opened 2018-01-10
Loose coupling of RiotApi

Oucema90:
#468

PR #510, opened 2018-01-17
Refactored Async Methods

Kirides:
Added .ConfigureAwait(false) to async Methods. 
GetAsync() now only gets the ResponseHeaders.

- By adding .ConfigureAwait(false) to all the async Methods, the Json-Serialization is automatically run on the ThreadPool. (Also kinda helps against .Result-/.Wait-Deadlocks, by not waiting for the blocked context)
- GetAsync only needs the Headers, because it's being followed up by GetResponseContentAsync. This way there is no MemoryStream created when reading the response content.

### DIFF
--- a/RiotSharp/Endpoints/ChampionEndpoint/ChampionEndpoint.cs
+++ b/RiotSharp/Endpoints/ChampionEndpoint/ChampionEndpoint.cs
@@ -30,9 +30,8 @@ namespace RiotSharp.Endpoints.ChampionEndpoint
         public async Task<List<Champion>> GetChampionsAsync(Region region, bool freeToPlay = false)
         {
             var json = await _requester.CreateGetRequestAsync(PlatformRootUrl + ChampionsUrl, region,
-                new List<string> { $"freeToPlay={freeToPlay.ToString().ToLower()}" });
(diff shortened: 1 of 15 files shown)


Kirides:
> Also, would these changes allow us to remove the synchronous versions of those methods?

@BenFradet Yes, since these changes would prevent a library deadlock, people could use ".Result" to receive the results in a synchronous fashion.


refactored the other 2 Apis as well. And reverted a little change (silly me).

JanOuborny:
Looks great 👍

However, I would wait with merging until #508 is done and then rebase your branch onto the latest version and apply the changes again.

BenFradet:
@Kirides it's good to rebase now that we've merged #508 👍

Kirides:
I adjusted the changes (phew, quite the rewrite of the library though)

BenFradet:
mmmh something must have gone wrong during your rebase because I'm seeing the commit for #508 in this PR.

Feel free to open a  new PR if you don't manage to clean the git history 👍

Kirides:
Okay - looks like i got it this time.

BenFradet:
Thanks, merging! 👍

Issue #500, opened 2017-12-17
Add third-party-code-v3 Endpoint


jwrichard:
Riot added a new endpoint used for 3rd parties to verify a summoner. Used to fix the old "Change your rune page name to xyz to verify".

Blog Post: https://discussion.developer.riotgames.com/questions/4269/verification-system.html
API Doc: https://developer.riotgames.com/api-methods/#third-party-code-v3/GET_getThirdPartyCodeBySummonerId

BenFradet:
Good idea 👍

jwrichard:
@BenFradet @Oucema90 Thank you 🙏

PR #512, opened 2018-01-19
Add third-party-code-v3 Endpoint #500

Oucema90:
ThirdParty Endpoint implementation.

### DIFF
--- a/RiotSharp.Test/RiotApiTest.cs
+++ b/RiotSharp.Test/RiotApiTest.cs
@@ -389,7 +389,7 @@ namespace RiotSharp.Test
         }
         #endregion
 
-        #region Thrid Party Tests
+        #region Third Party Tests
         [TestMethod]
         [TestCategory("RiotApi")]
         public void GetThirdPartyCode_Test()
--- a/RiotSharp.Test/RiotApiTest.cs
+++ b/RiotSharp.Test/RiotApiTest.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Net;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using RiotSharp.Endpoints.MatchEndpoint.Enums;
+using RiotSharp.Misc;
 
 namespace RiotSharp.Test
 {
@@ -386,6 +387,36 @@ namespace RiotSharp.Test
                 Assert.IsTrue(totalChampionMasteryScore > -1);
             });
         }
+        #endregion
+
+        #region Thrid Party Tests
+
+        [TestMethod]
+        [TestCategory("RiotApi")]
+        public void GetThirdPartyCode_Test()
+        {
+            EnsureCredibility(() =>
+            {
+                var code = Api.ThirdParty.GetThirdPartyCodeBySummonerId(Summoner1And2Region,
+                    Summoner1Id);
+
+                Assert.AreEqual(RiotApiTestBase.ThirdPartyCode, code);
+            });
+        }
+
+        [TestMethod]
+        [TestCategory("RiotApi"), TestCategory("Async")]
+        public void GetThirdPartyCodeAsync_Test()
+        {
+            EnsureCredibility(() =>
+            {
+                var code = Api.ThirdParty.GetThirdPartyCodeBySummonerIdAsync(Summoner1And2Region,
+                    Summoner1Id);
+
+                Assert.AreEqual(RiotApiTestBase.ThirdPartyCode, code.Result);
+            });
+        }
+
         #endregion
     }
 }
--- a/RiotSharp.Test/RiotApiTest.cs
+++ b/RiotSharp.Test/RiotApiTest.cs
@@ -390,7 +390,6 @@ namespace RiotSharp.Test
         #endregion
 
         #region Thrid Party Tests
-
         [TestMethod]
         [TestCategory("RiotApi")]
         public void GetThirdPartyCode_Test()
@@ -416,7 +415,6 @@ namespace RiotSharp.Test
                 Assert.AreEqual(RiotApiTestBase.ThirdPartyCode, code.Result);
             });
         }
-
         #endregion
     }
 }
--- a/RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
+++ b/RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
@@ -17,7 +17,6 @@ namespace RiotSharp.Endpoints.ThirdPartEndpoint
             _requester = requester;
         }
 
-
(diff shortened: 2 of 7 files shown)


RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
```diff
@@ -0,0 +1,34 @@
...
+        private const string ThirdPartyBySummonerUrl = "/by-summoner/{0}";
+
+        private readonly IRateLimitedRequester _requester;
+
+        public ThirdPartyEndpoint(IRateLimitedRequester requester)
+        {
+            _requester = requester;
+        }
+
+
```
BenFradet: nit: empty line

RiotSharp.Test/RiotApiTest.cs
```diff
@@ -744,5 +745,35 @@ public void GetTotalChampionMasteryScoreAsync_Test()
...
+            EnsureCredibility(() =>
+            {
+                var code = Api.ThirdParty.GetThirdPartyCodeBySummonerIdAsync(Summoner1And2Region,
+                    Summoner1Id);
+
+                Assert.AreEqual(RiotApiTestBase.ThirdPartyCode, code.Result);
+            });
+        }
+
+        #endregion
```
BenFradet: Do we want to keep those integration tests or just the mocks?

Oucema90:
@BenFradet Thanks for the positive feedback. PR updated.

RiotSharp.Test/RiotApiTest.cs
Oucema90: Regarding this question : in my opinion we should keep both. Since integration test will cover our pieces and shows if it interacts well with this last; while moq tests intend to check if the logic implemented/ refactored works as expected or not.

RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
```diff
@@ -0,0 +1,33 @@
...
+                region);
+        }
+
+        public async Task<string> GetThirdPartyCodeBySummonerIdAsync(Region region, long summonerId)
+        {
+            return await _requester.CreateGetRequestAsync(
+                string.Format(ThirdPartyRootUrl + ThirdPartyBySummonerUrl, summonerId),region).ConfigureAwait(false);
+        }
+    }
+}
```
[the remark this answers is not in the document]
Oucema90: I removed it unintentionally.

Oucema90:
@JanOuborny Thanks for the feedback.

PR updated.

RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
```diff
@@ -19,15 +19,16 @@ public ThirdPartyEndpoint(IRateLimitedRequester requester)
 
         public string GetThirdPartyCodeBySummonerId(Region region, long summonerId)
         {
-            return _requester.CreateGetRequest(ThirdPartyRootUrl + 
-                string.Format(ThirdPartyBySummonerUrl, summonerId),
-                region);
+            return _requester
+                .CreateGetRequest(ThirdPartyRootUrl + string.Format(ThirdPartyBySummonerUrl, summonerId), region)
+                .Replace("\"", "");
```
JanOuborny: I think the `Replace` is not needed

RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
```diff
@@ -19,15 +19,16 @@ public ThirdPartyEndpoint(IRateLimitedRequester requester)
...
         }
 
         public async Task<string> GetThirdPartyCodeBySummonerIdAsync(Region region, long summonerId)
         {
-            return await _requester.CreateGetRequestAsync(
-                string.Format(ThirdPartyRootUrl + ThirdPartyBySummonerUrl, summonerId),region).ConfigureAwait(false);
+            return (await _requester
+                .CreateGetRequestAsync(string.Format(ThirdPartyRootUrl + ThirdPartyBySummonerUrl, summonerId), region)
+                .ConfigureAwait(false)).Replace("\"", "");
```
JanOuborny: same here

RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
Oucema90: I used replace in order to improve readability you mentioned here  

> You might want to change this into a string literal to improve readability 

If we remove if our two new test will fail for this case. 
```
Assert.AreEqual failed. Expected:<AD irelia (AS)>. Actual:<"AD irelia (AS)">
```

MingweiSamuel: What if somebody has a quote character in their 3d party code?

RiotSharp.Test/RiotApiTest.cs
```diff
@@ -744,5 +745,33 @@ public void GetTotalChampionMasteryScoreAsync_Test()
             });
         }
         #endregion
+
+        #region Thrid Party Tests
```
JanOuborny: typo

RiotSharp/Endpoints/ThirdPartEndpoint/ThirdPartyEndpoint.cs
JanOuborny: Riot returns a string with quotation marks, e.g. `"AD irelia (AS)"` . We could leave it as it is. However, it might be confusing to use. Just remove the quotation marks at the beginning and at the end and make sure it's documented in the interface method.

JanOuborny: A simple solution would be:
```
	var rawResponse = _requester.CreateGetRequest(ThirdPartyRootUrl + 
		string.Format(ThirdPartyBySummonerUrl, summonerId), region);
	return rawResponse.Substring(1, rawResponse.Length - 2);
```

BenFradet: we might want to check the prefix and suffix of the string before removing characters

MingweiSamuel: The response content is still technically JSON (a json string literal) so you should still be able to use JSON.net to parse it.

Double quotes will end up as `\"` and unicode may (or may not be) encoded as `\u00f8` and still be valid JSON.

BenFradet:
Merging, thanks a lot! 👍

Issue #468, opened 2017-09-16
Loose coupling of RiotApi 


JanOuborny:
From a maintenance and DI view, it makes sense to split the `RiotApi` in to different endpoints, which are then combined back together in the `RiotApi` class.

I propose following solution architecture:

* `RiotSharp.Endpoints.Interfaces`
  - `ISpectatorEndpoint`
  - `IRunesEndpoint`
  - `...`
* `RiotSharp.Endpoints`
    - `SpectatorEndpoint`
    - `RunesEndpoint`
    - `...`

In the `RiotApi` class the endpoints would be referenced like this:
``` c#
private ISpectatorEndpoint spectatorEndpoint = new SpectatorEndpoint(...);
public ISpectatorEndpoint Spectator => spectatorEndpoint;
```

Consumption of the endpoints would look like this:
``` c#
var api = RiotApi.GetInstance();
api.Spectator.GetCurrentGame(region, id);
```

Shidesu:
For 3.1 i think :')
It will be nice and easier to maintain -> but a bit long to put in place.

JanOuborny:
That's not really that hard. I would say in two to three hours it's done.

BenFradet:
yup I think the most urgent stuff is to support the v3 api, this is a bit longer term 👍

PR #515, opened 2018-01-20
Completely refactored the StaticApi. (#514)

### LINKED ISSUE #514: Remove StaticRiotApi and include it as an endpoint

Kirides:
- The StaticApi is now a property on the RiotApi (api.Static)
- All Endpoints of the StaticApi are now properties of the IStaticDataEndpoint (api.Static.Champion, ...)
- The StaticDataEndpoint fully supports dependency injection.
- Every static Endpoint has its own configuration (SlidingExpirationTime, Cache, Requester)
     - the default StaticEndpointProvider configures all endpoints to use the same requester and cache
- All static Endpoints derive from IStaticEndpoint (empty interface)

----
Changes were made in order to fix #514.

I might have gone quite overboard with this :/


RiotSharp/Endpoints/Interfaces/Static/IStaticDataEndpoint.cs
```diff
@@ -0,0 +1,18 @@
+﻿using RiotSharp.Endpoints.Interfaces.Static;
+
+namespace RiotSharp.Endpoints.Interfaces.Static
+{
+    public interface IStaticDataEndpoint
+    {
+        IStaticChampionEndpoint Champion { get; }
```
JanOuborny: Remove all other `IStaticEndpoint` and put all methods in here

BenFradet:
@JanOuborny I disagree, I think this approach makes it way clearer and easier to navigate.

JanOuborny:
@BenFradet That's true. However, now you are misusing the concept of endpoints.

BenFradet:
How so? For me `lol/static-data/v3/champions` is one endpoint and the full list of the static endpoints is an API. Furthermore, it makes sense to group endpoints logically (e.g. `lol/static-data/v3/champions` and `/lol/static-data/v3/champions/{id}`) as has been done in this PR.

Kirides:
I thought of it like the regular API, where every endpoint is a separate property. 

Where lol/{endpoint} is the root class
With properties for each endpoint.

In the way that #514 implied to add the static API as an endpoint  one could choose between
- Api.Static.Champion
- Api.ChampionStatic

At least the way i read into it.
But I'd rather have all endpoints separated and group them under the IStaticData instead of having another huge all-in-one class which would be no better than StaticRiotApi IMO.

JanOuborny:
Each section in Riot's full api reference is an individual API as far as I know.
That would result in the following URL format: /lol/{api}/{endpoint}/{specific-operation}

Kirides:
But this would mean that the static API should be a separate API class (StaticRiotApi) and this does not match the mentioned Issue #514.

JanOuborny:
That's true, because in general the API and endpoint naming is misused in this project. Especially because I introduced it in #468 and by using a `RiotApi` class to wrap all apis.

Kirides:
Well the way I implemented the separate endpoints (not the static meta endpoint) should match the proper use of an endpoint (being a endpoint providing an API).

The issue i see is, that the usage of the StaticData property is abusing the naming.

But from a consumers point, this way feels more natural

RiotSharp/Endpoints/Interfaces/Static/IStaticDataEndpoint.cs
```diff
@@ -0,0 +1,18 @@
+﻿using RiotSharp.Endpoints.Interfaces.Static;
+
+namespace RiotSharp.Endpoints.Interfaces.Static
+{
+    public interface IStaticDataEndpoint
+    {
```
JanOuborny: You might rename this to `IStaticDataEndpointCollection` or `IStaticDataEndpoints`, so it's distinguishable from the actual endpoints.

RiotSharp/Endpoints/Interfaces/Static/IStaticDataEndpoint.cs
```diff
@@ -0,0 +1,18 @@
+﻿using RiotSharp.Endpoints.Interfaces.Static;
+
+namespace RiotSharp.Endpoints.Interfaces.Static
+{
+    public interface IStaticDataEndpoint
```
JanOuborny: You might rename this to `IStaticDataEndpointCollection` or `IStaticDataEndpoints`, so it's distinguishable from the actual endpoints.

Kirides: Since it is no "real" collection, the second one looks more promising.

@BenFradet what is your opinion on the change from `IStaticDataEndpoint` to IStaticDataEndpoints`?

BenFradet: let's go with `IStaticDataEndpoints` 👍

Kirides:
Rebased changes on latest develop-branch and changed naming of `IStaticDataEndpoint`

btw. someone let this  code go through... *cough*
`Endpoints.ThirdPartyEndpoint`
```cs
ThirdPartyEndpoint : IThirdPartEndpoint // Interface says "Part" instead of Party
... 
// Method is synchronous
public string GetThirdPartyCodeBySummonerId(Region region, long summonerId)
{
    var response = _requester
            .CreateGetRequest(ThirdPartyRootUrl + string.Format(ThirdPartyBySummonerUrl, summonerId), region);
    return response.Substring(1, response.Length - 2);
}
```

BenFradet:
arf my bad missed the above during the review

@Oucema90 could you fix the above?

Oucema90:
@Kirides It was my bad. I have missed it too.

@BenFradet PR created #518.

Kirides:
Merged chages from @Oucema90 

(Can not access PC for a few days, merged using github)

BenFradet:
This is good to go from me, wdyt @janouborny?

BryanHitchcock:
Could we get this merged, please?

BryanHitchcock:
@BenFradet @JanOuborny Could we get this merged soon, please?

Kirides:
Fixed the build issues and added the IStaticEndpointProvider to the AspNetCore Extension

BenFradet:
is this now good to go @JanOuborny ?

JanOuborny:
@BenFradet Yes, it looks good.

BenFradet:
Merging, thanks!

Issue #501, opened 2017-12-17
Add a top-level Cache namespace


BenFradet:
`RiotSharp.Cache` should contain:

- [Cache](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/Cache.cs)
- [ICache](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/ICache.cs)
- [PassThroughCache](https://github.com/BenFradet/RiotSharp/blob/develop/RiotSharp/PassThroughCache.cs)
- [FileCache](https://github.com/Talha-T/RiotSharp/blob/86f92adb0b92a35b9831858d2c8694a00fa834a2/RiotSharp/FileCache.cs)
- [CacheData](https://github.com/Talha-T/RiotSharp/blob/86f92adb0b92a35b9831858d2c8694a00fa834a2/RiotSharp/CacheData.cs)

PR #516, opened 2018-01-22
Add a top-level Cache namespace #501

Oucema90:
Move Cache classes to the new namespace.


Oucema90:
Since it regroups multiples Cache; plural form is more suitable here. Also there will be a confusion if we use Cache as name space and Cache as class.

JanOuborny:
What about renaming it to `Caching`? This naming is also used by Microsoft in AspNetCore.

Oucema90:
@JanOuborny PR updated.

BenFradet:
@Oucema90 could you rebase against develop? I'll merge right after that 👍

Oucema90:
@BenFradet PR updated.

PR #522, opened 2018-02-03
2018 season converter

jeuxjeux20:
Forgot there was a converter, tweaked it :p


jeuxjeux20:
But on a side note, all these converters can be removed ! 

Newtonsoft.Json automatically transforms numbers into enums.

PR #541, opened 2018-04-10
Setting RiotSharp references CopyLocal to false.

Trigger-edge:
Set all NetStandard.Library references CopyLocal to false.


BenFradet:
@Trigger-edge please pr against develop 👍

PR #548, opened 2018-05-15
Change gameDuration from milliseconds to seconds

Vertana:
Changes gameDuration JSON property from milliseconds to seconds to comply with League of Legends API at the time of this patch. Also cleans up name from "TimeSpanConverterFromS" to "TimeSpanConverterFromSeconds" to match the millisecond converter name.

This is from https://developer.riotgames.com/api-methods/#match-v3/GET_getMatch

You can see the gameDuration property returns a long. This value is returned in seconds. This has been tested and verified locally.


BenFradet:
Thanks, merged!

PR #554, opened 2018-06-14
Remove Runes & Masteries Endpoints #553

### LINKED ISSUE #553: Remove Runes&Masteries Endpoints?
Following #552 

We have tests which are falling because runes&masteries api was removed due to massive rework a while ago.
```
GetRunePagesAsync_ExistingSummonerId_HasRunePages
GetRunePagesAsync_InvalidSummonerId_ThrowsResouceNotFound
GetMasteryPagesAsync_ExistingSummonerId_HasMasteryPages
GetMasteryPagesAsync_InvalidSummonerId_ThrowsResouceNotFound
```
From https://discussion.developer.riotgames.com/articles/3779/runes-reforged-changes-in-the-riot-games-api.html
>this means we’ll be deprecating runes-v3 and masteries-v3 with no replacement. Runes Reforged data will be included in both the active game and match data.

I suggest we remove these Endpoints as well as tests. I will submit PR if you are ok.

CurlyBraceTT:
Fixes #553  

Removed Runes Endpoint and Tests
Removed Masteries Endpoint and Tests
Removed unused using directives in modified files


CurlyBraceTT:
@BenFradet Actually my bad here, accidentally committed appsettings.json file. You might want to revert this file back, and add maybe it to the git ignore.

BenFradet:
We have a script that puts my api key to run the tests https://github.com/BenFradet/RiotSharp/blob/develop/appveyor.yml#L52 so we can't add it to the gitignore unfortunately.

What I usually do is `git update-index --assume-unchanged` http://gitready.com/intermediate/2009/02/18/temporarily-ignoring-files.html

CurlyBraceTT:
@BenFradet Thanks, that works :thumbsup:

Issue #580, opened 2018-10-10
Add free champion rotation endpoint


JanOuborny:
Ref: https://developer.riotgames.com/api-methods/#champion-v3/GET_getChampionInfo

JanOuborny:
Resolved in #586.

PR #586, opened 2018-10-14
Free Rotation

SJenn:
Tried to address this issue:
https://github.com/BenFradet/RiotSharp/issues/580
Please let me know if this solution is acceptable.


RiotSharp/Endpoints/ChampionRotationEndpoint/ChampionRotationEndpoint.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <inheritdoc />
+        public async Task<ChampionRotation> GetChampionRotationAsync(Region region)
+        {
+            var json = await _requester.CreateGetRequestAsync(PlatformRootUrl + ChampionsUrl, region
+                ).ConfigureAwait(false);
+            return JsonConvert.DeserializeObject<ChampionRotation>(json);
+        }
+    }
+}
```
JanOuborny: Missing newline

SJenn: Attempted to edit it using github online editor. Let me know if that fixed it.

Kirides:
Technically the Endpoint/API is still the "ChampionEndpoint" from an API perspective, renaming from IChampionEndpoint to IChampionRotationEndpoint does not seem like a good idea to me.

For example: What's going to happen when the ChampionEndpoint gets another Method independant from the ChampionRotation?
- ... another Interface needs to be created and implemented
- .. or the interface/Impl. is renamed again

SJenn:
What you are basically saying is that instead of making it as a new endpoint I should just make it as a new method for the ChampionEndpoint and it's interface? If that's so, I'll let the other people that reviewed this change to comment on it. If I misunderstood, feel free to say so.

JanOuborny:
I agree. The plan was to follow the structure of the full api reference. Right now the champion rotation is a operation of the champion endpoint. Therefore, it should be added to the `IChampionEndpoint`.

SJenn:
I attempted to change it according to the suggestions here. Feel free to let me know if it's the way you imagined now.

PR #589, opened 2018-10-21
Use X-Riot-Token Header instead of api_key querystring

Kirides:
Refactored RequesterBase so it uses "X-Riot-Token" HTTP-Header instead of "api_key" querystring.

Metadata belongs to the HTTP Requestheaders and it overall seems wrong to "put authorization into a url" also improves possible future caching logic, where the cache-key might be combined from the querystring

### DIFF
--- a/RiotSharp/Http/RateLimitedRequester.cs
+++ b/RiotSharp/Http/RateLimitedRequester.cs
@@ -30,34 +30,24 @@ namespace RiotSharp.Http
         #region Public Methods
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


Kirides:
@JanOuborny I now removed these functions and additionaly reduced some clutter. in RateLimitedRequester.

While reducing the redundant code, i came across the `Task<bool> CreatePutRequestAsync`-Method, which's usage is not clear to me. When it sends the request which in turn yields a `!response.IsSuccessStatusCode` the code throws an exception, which makes the return type `bool` useless, because it's only ever `true` when there is no issue

Issue #588, opened 2018-10-19
Remove old Runes / Masteries from CurrentGame Endpoint and add Runes Reforged[Perks] + StaticApi


lorexzer0:
- [x] The old runes / masteries system is removed from league so they should be removed from the CurrentGame.Endpoint, altough u can still find data about them in older matches so they should stay in the Match.Endpoint. 
- [x] Add Perks to the CurrentGame.Endpoint so the runes can be aquired.
- [x] Use "[ddragon](http://ddragon.leagueoflegends.com/cdn/8.20.1/data/en_US/champion.json)" for static data endpoint. More about that here: [Static Data](https://developer.riotgames.com/static-data.html)

cornerstone-programming:
Regarding second point "ddragon"

I m wondering if riotsharp is just a wrapper for riotapi or if ddragon will be part as well.

JanOuborny:
The latest build already contains a wrapper for the ddragon data, which was provided in the static data endpoint.

PR #596, opened 2018-12-10
Update spectator and match endpoint to match the current API state

JanOuborny:
- Remove runes and masteries from featured games endpoint #588 
- Add perk stats to ParticipantStats of match endpoint
- Update spectator endpoint
- Improve unit tests


RiotSharp/Endpoints/SpectatorEndpoint/FeaturedGame.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// The ID of the game
+        /// </summary>
+        [JsonProperty("gameId")]
+        public long GameId { get; set; }
+
+        /// <summary>
+        /// The amount of time in seconds that has passed since the game started
+        /// </summary>
+        [JsonProperty("gameLength")]
+        public long GameLength { get; set; }
```
BenFradet: could we turn that into a duration?

PR #622, opened 2019-04-26
Adjust league (Fixes #619, #621 & #623)

### LINKED ISSUE #619: PlayerOrTeamName is Null
League.GetLeaguePositionsAsync

image

### LINKED ISSUE #621: Update league endpoint
Add
- `/lol/league/v4/entries/by-summoner/{encryptedSummonerId}`
- `/lol/league/v4/entries/{queue}/{tier}/{division}`
- `/lol/league/v4/grandmasterleagues/by-queue/{queue}`
- `/lol/league/v4/leagues/{leagueId}`

Remove:

- `GetLeaguePositionsAsync`

Ref:
https://developer.riotgames.com/api-methods/#league-v4/GET_getLeagueById

### LINKED ISSUE #623: GetChallengerLeagueAsync has incorrect modal
When im calling `await api.League.GetChallengerLeagueAsync(Region.Euw, Queue.RankedSolo5x5);` the enties modal looks like this: 

image

As you can see `leagueName` is null for all entries. I think this is because the Riot api defines this property as `summonerName` and also `summonerId`, which is missing. Changing the region or queue doen't change the result.

I think that this might be a bug since the `LeaguePosition` model doesn't corresponds with the RiotApi model.

kirides:
Should close #619, fix #621 and fix #623

- Removed GetLeagePositionsAsync
- Added four new methods for the endpoints defined in #621
- ~~fixed IServiceCollectionExtensions for AspNetCore~~
- TierConverter now uses Enum.TryParse(string, bool ignoreCase, out enum)
- Base-Classed LeagueEntry (LeaguePosition replacement) with LeagueItem. (Returned from LeagueList-returning endpoints

### DIFF
--- a/RiotSharp.Test/RiotApiTest.cs
+++ b/RiotSharp.Test/RiotApiTest.cs
@@ -23,7 +23,7 @@ namespace RiotSharp.Test
         {
             EnsureCredibility(() =>
             {
-                var summoner = Api.Summoner.GetSummonerBySummonerIdAsync(Summoner1And2Region, 
+                var summoner = Api.Summoner.GetSummonerBySummonerIdAsync(Summoner1And2Region,
                     Summoner1Id);
 
                 Assert.AreEqual(Summoner1Name, summoner.Result.Name);
@@ -37,7 +37,7 @@ namespace RiotSharp.Test
         {
             EnsureCredibility(() =>
             {
-                var summoner = Api.Summoner.GetSummonerByAccountIdAsync(Summoner1And2Region, 
+                var summoner = Api.Summoner.GetSummonerByAccountIdAsync(Summoner1And2Region,
                     Summoner1AccountId);
 
                 Assert.AreEqual(Summoner1Name, summoner.Result.Name);
@@ -50,7 +50,7 @@ namespace RiotSharp.Test
         {
             EnsureCredibility(() =>
             {
-                var summoner = Api.Summoner.GetSummonerByNameAsync(Summoner1And2Region, 
+                var summoner = Api.Summoner.GetSummonerByNameAsync(Summoner1And2Region,
                     Summoner1Name);
 
                 Assert.AreEqual(Summoner1Id, summoner.Result.Id);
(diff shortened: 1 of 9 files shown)


RiotSharp.AspNetCore/IServiceCollectionExtension.cs
```diff
@@ -33,6 +33,8 @@ public static IServiceCollection AddRiotSharp(this IServiceCollection serviceCol
             {
                 var rateLimitedRequester = new RateLimitedRequester(riotSharpOptions.RiotApi.ApiKey,
                     riotSharpOptions.RiotApi.RateLimits);
+                var nonLimitingRequester = new Requester(riotSharpOptions.RiotApi.ApiKey);
```
BenFradet: why do you need it?

BenFradet (commented):
LGTM, I just have one q :+1:

RiotSharp.AspNetCore/IServiceCollectionExtension.cs
kirides: Someone added the status endpoint to the API, which lead to the API taking another rate limiter in it's constructor.

IDK if its really necessary, but if the status API is not limited, reusing the rate limited requester would lead to rate limits potentially being fired without hitting a limited endpoint.

JanOuborny: The problem here is: Right now every requester creates his own `HttpClient` instance. However, the `HttpClient` is threadsafe and designed for concurrent requests, therefore we should reuse the same instance for every requester.  When I migrated the status API, I intentionally left the build error there to fix it properly in a later commit.

kirides: @JanOuborny Alright, removed that part from the PR :)

JanOuborny:
Looks good 👍 
Just a little thing: Can you add the supported values disclaimer [here](https://github.com/BenFradet/RiotSharp/blob/4c721e512debb8a4bd45b35ba00fef99528618e4/RiotSharp/Endpoints/Interfaces/ILeagueEndpoint.cs#L17)?

kirides:
Done

BenFradet:
are we good to go here @JanOuborny ?

JanOuborny (approved):
I think so @BenFradet

PR #633, opened 2019-09-24
Thread-safe RateLimitedRequester

lezzi:
`RateLimitedRequester` is not thread-safe at the moment and it throws error when several requests for the same region are made concurrently right after class creation. This bug was found in production.

### DIFF
--- a/RiotSharp/Http/RateLimitedRequester.cs
+++ b/RiotSharp/Http/RateLimitedRequester.cs
@@ -5,6 +5,7 @@ using System.Text;
 using System.Threading.Tasks;
 using RiotSharp.Misc;
 using System;
+using System.Collections.Concurrent;
 
 namespace RiotSharp.Http
 {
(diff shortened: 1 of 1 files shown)


MingweiSamuel (approved):
Looks like I missed that, good catch. Can't guarantee there aren't any other synchronization issues but this change can't hurt

lezzi:
From my experience I can say that it seems to be working stable in concurrent scenarios, but this particular situation was an edge case - it happened after several months of using the same code in prod and caused dictionary to enter corrupted state. Restart of course helped, but it can happen randomly after any app restart.

BenFradet (approved):
Thanks :+1:

matheusavi:
This helped me so much!
Thanks!

svstoichkov:
Please merge this ASAP. Its causing a lot of issues for us in production

sguryev:
Please listen to @svstoichkov. Our production is broken as well now. We have wasted more than 3 days to find the issue

BenFradet:
Sorry, I was waiting for @JanOuborny 's review but it never came :+1:.

Issue #650, opened 2020-02-04
FileCache not working


xXLAOKOONXx:
I discovered multiple issues around the FileCache:

Path.Combine can not work with '/' at first position
Cache folder not created if not existing
Load<T>/Get<T> do not use GetPath

PR #651, opened 2020-02-04
Fix FileCache

xXLAOKOONXx:
Fixing FileCache class
See #650


xXLAOKOONXx:
#652 Solves the issue and takes additional factors into account.

PR #652, opened 2020-02-06
Fixed FileCache implementation

SergeantSerk:
Added my fixed implementation of FileCache where a cryptographic hash function (SHA1) is used instead of HashCode so the hash persists over different processes.

Currently, using the wrapper twice over different process runtimes yields different file names for the same subject, whereas hashing it using a cryptographic function yields the same name, also respecting the TTL.

One thing to consider might be using SHA256 instead of SHA1/MD5 since collisions have been found for the latter mentioned hash functions, though SHA256 has longer filenames which may be a problem for filesystems with file path limit.

### DIFF
--- a/RiotSharp/Caching/FileCache.cs
+++ b/RiotSharp/Caching/FileCache.cs
@@ -12,13 +12,14 @@ namespace RiotSharp.Caching
     /// <seealso cref="RiotSharp.Caching.ICache" />
     public class FileCache : ICache
     {
-        private static string _directory;
+        private string _directory;
+        private bool _hashKeys;
 
         /// <summary>
         /// Create file cache instance
         /// </summary>
         /// <param name="dir">Directory for the cache to store in</param>
-        public FileCache(Uri directory)
+        public FileCache(Uri directory, bool hashKeys = false)
         {
             if (directory == null)
             {
@@ -40,6 +41,7 @@ namespace RiotSharp.Caching
             }
 
             Directory.CreateDirectory(_directory);
+            _hashKeys = hashKeys;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


xXLAOKOONXx:
I like your approach! (Will close my pull request linking to yours)
I figured out it might be a good idea to offer a constructor with full directory path, e.g. D:/myLoLApp/cache
This way you might be able to counter your concerns about long file names and you have more freedom overall.
@SergeantSerk 

question out of interest: Is there an advantage in performance by using hash values? Didn't found anything about such a thing yet.

xXLAOKOONXx:
Fixes #650

SergeantSerk:
@xXLaokoonXx 
> I figured out it might be a good idea to offer a constructor with full directory path, e.g. D:/myLoLApp/cache

Good idea. After checking your PR, I've noticed what you meant and used `System.Uri` to implement that further. I only tried relative and absolute in basic terms (`..\cache` and `C:\Test\cache` directory paths respectively) which worked.

RE: the long file problem, I might stick with SHA1 since the class in question isn't concerned with being cryptographically secure as there is nothing to be secured, and SHA256 and upwards is relatively computationally intensive, without downgrading to MD5. Anyone with concerns RE this can notify me/modify as such.

> question out of interest: Is there an advantage in performance by using hash values? Didn't found anything about such a thing yet.

Calculating hashes is pretty computationally intensive so it depends on your use case. Collections like `HashSet<>` use hashes to store data, which allows for quick retrieval of data (O(1) time complexity) at the cost of spending some time to initially insert the data by hashing it. Though, there is no performance advantage of it here (within FileCache) since we're just naming files, but using hashes should guarantee uniqueness (the reason why MD5/SHA1 is considered broken for cryptographic applications is because of such collisions).

JanOuborny:
Technically, hashing of the cache keys to create a file name isn't necessary, because the cache keys should be unique by definition. I don't really remember why the old implementation used it.

SergeantSerk:
You're right, I must've gotten carried away by the `GetHashCode` in the old implementation, though if there is a need for uniform file name length, then this is achieved through hashing, otherwise it is useless here and a waste of computing power.

I've updated the constructor and `GetPath` method as such, to allow for the developer to choose either of the options, defaulting to no hashing of keys. Also, there was an issue of the directory path not being respected per instance due to the directory variable being static. This is addressed.

BenFradet (approved):

JanOuborny (approved):

Issue #654, opened 2020-02-25
Cache not working


banitt:
In this example
```
var cache = new Cache();
cache.Add("data", "test", TimeSpan.FromMilliseconds(100));

while(true)
{
    var item = cache.Get<string,string>("data");
    Console.WriteLine(item.ToString());
}
```

item in cache should be available for 100 milliseconds, but instead it is always available.

xXLAOKOONXx:
I am not that confident in this caching.
I found a code snipped that refreshes the expiry duration every time you access it:

Call of .Viewed() inside the Get function
https://github.com/BenFradet/RiotSharp/blob/7ae756c8ce7eb1c700739f85f8b5fc8fd8521723/RiotSharp/Caching/Cache.cs#L51
Code inside .Viewed()
https://github.com/BenFradet/RiotSharp/blob/7ae756c8ce7eb1c700739f85f8b5fc8fd8521723/RiotSharp/Caching/Cache.cs#L282

This behavior makes kinda sense, so I guess it is intended.
@banitt may you check if caching works if you delay for a second before accessing the data in an endless loop?

banitt:
@xXLaokoonXx 
Oh, I thought item should be available up to 100 milliseconds.

xXLAOKOONXx:
FileCache works as you expect it to work.
As Cache and FileCache claim to do the same with different backend implementation it might be a good point to align both approaches regarding expiry.
In the same change Cache could implement CacheData.
@JanOuborny or @BenFradet your thoughts?

BenFradet:
>As Cache and FileCache claim to do the same with different backend implementation it might be a good point to align both approaches regarding expiry.

agreed :+1:

PR #655, opened 2020-02-25
Align Cache with FileCache implementation

xXLAOKOONXx:
Related to issue #654 .

- Aligned understanding of expiry duration (Now expires at creation + timespan even if the data was revisited)
- Used CacheData instead of private classes
- Renamed and extracted functionalities based on implementation in FileCache (Save and Load)


BenFradet (approved):
Thanks :+1:

PR #656, opened 2020-03-21
Fix AspNetCore registration

Nicolas-Ding:
The RiotApi constructor now requires a Requester as argument, otherwise the project fails to compile


BenFradet (approved):
Thanks 👍

JanOuborny (approved):

PR #657, opened 2020-03-23
Fix some XML comments for publicly visible methods as they generate warnings

Nicolas-Ding:
When building the project, I get the following warnings : 

image


BenFradet (approved):
Thanks a lot :+1:

PR #658, opened 2020-03-23
Fix some more CS1591

Nicolas-Ding:
When building the project, there are still a lot (around 110) of [CS 1591 warnings](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591).

This PR fixes some of them. 

Expect more PRs to come :)

image


Nicolas-Ding:
I'm using RiotSharp as a git submodule (because the nuget is not up to date) and it's kind of annoying to have all those warnings so I'm fixing them :)

BenFradet (approved):
Keep'em coming! :+1:

PR #659, opened 2020-03-24
Fix more CS1591

Nicolas-Ding:
Fix 31 more [CS 1591 warnings](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591) errors (82 -> 51)


BenFradet (approved):
Thanks :+1:

PR #661, opened 2020-04-28
Refactor AspNetCore project

JanOuborny:
- Added caching type
- Structured classes in namespaces

### DIFF
--- a/RiotSharp.AspNetCore/Options/ApiKeyOptions.cs
+++ b/RiotSharp.AspNetCore/Options/ApiKeyOptions.cs
@@ -10,7 +10,7 @@ namespace RiotSharp.AspNetCore.Options
     public class ApiKeyOptions
     {
         internal ApiKeyOptions() {
-            CacheType = CacheType.Internal;
+            CacheType = CacheType.PassThrough;
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


RiotSharp.AspNetCore/Options/ApiKeyOptions.cs
```diff
@@ -1,14 +1,17 @@
...
+namespace RiotSharp.AspNetCore.Options
 {
     /// <summary>
     /// Specifies the options for ApiKey
     /// </summary>
     public class ApiKeyOptions
     {
-        internal ApiKeyOptions() { }
+        internal ApiKeyOptions() {
+            CacheType = CacheType.Internal;
```
BenFradet: if internal is not recommended maybe we should use another as default?

BenFradet (approved):
LGTM, I just have one question around the default

RiotSharp.AspNetCore/Options/ApiKeyOptions.cs
Kiritsu: To be consistent with the rest of the lib, I'd put no cache by default.

Kiritsu:
Shouldn't you let the user, if they want to, use their own implementation of `ICache`?

RiotSharp.AspNetCore/Options/ApiKeyOptions.cs
JanOuborny: My bad. I wrongly thought that we use the internal implementation by default.

BenFradet (approved):

Issue #644, opened 2020-01-16
Refactor StaticDataEndpoints to be more user friendly


JanOuborny:
At first the endpoint should be renamed to DataDragon to match Riot's documentation naming (the old static data api was deprecated a long time ago).

Also getting data should be simplified, for example getting a champion by its id should look like this:
`var champion = await ddragon.Champions.GetByIdÀsync(version, id);`
instead of:
`var champion= ddragon.Champions.GetAllAsync(latestVersion).Result.Champions.Values.Single(x => x.Id == id);`

Endpoints to be refactored:
- [ ] Champion
- [ ] Item
- [ ] LanguageStrings
- [ ] Map
- [ ] Mastery
- [ ] ProfileIcons
- [ ] Realm
- [ ] ReforgedRunes
- [ ] Rune
- [ ] Champion
- [ ] SummonerSpell
- [ ] TarballLinks
- [ ] Version

Capm96:
Hey @JanOuborny is this still up for grabs? I'd like to help out.

Thanks!

JanOuborny:
Yes, it is still up for grabs. Go for it!

PR #670, opened 2020-08-09
Feature/644 refactor static data endpoints

Capm96:
Following #644 , renamed the StaticDataEndpoints class to DataDragonEndpoints. Now the user will call RiotApi.DataDragon for the static endpoints. 

Also added to the DataDragon.Champions a method to get a champion by its id.

### DIFF
--- a/RiotSharp.Test/RiotApiTest.cs
+++ b/RiotSharp.Test/RiotApiTest.cs
@@ -139,7 +139,7 @@ namespace RiotSharp.Test
         {
             EnsureCredibility(() =>
             {
-                var leagues = Api.League.GetLeagueGrandmastersByQueueAsync(RiotApiTestBase.SummonersRegion, RiotSharp.Misc.Queue.RankedSolo5x5);
+                 var leagues = Api.League.GetLeagueGrandmastersByQueueAsync(RiotApiTestBase.SummonersRegion, RiotSharp.Misc.Queue.RankedSolo5x5);
 
                 Assert.IsTrue(leagues.Result.Queue != null);
             });
--- a/RiotSharp.Test/RiotApiTest.cs
+++ b/RiotSharp.Test/RiotApiTest.cs
@@ -139,7 +139,7 @@ namespace RiotSharp.Test
         {
             EnsureCredibility(() =>
             {
-                 var leagues = Api.League.GetLeagueGrandmastersByQueueAsync(RiotApiTestBase.SummonersRegion, RiotSharp.Misc.Queue.RankedSolo5x5);
+                var leagues = Api.League.GetLeagueGrandmastersByQueueAsync(RiotApiTestBase.SummonersRegion, RiotSharp.Misc.Queue.RankedSolo5x5);
 
                 Assert.IsTrue(leagues.Result.Queue != null);
             });
--- a/RiotSharp/Endpoints/Interfaces/Static/IDataDragonEndpoints.cs
+++ b/RiotSharp/Endpoints/Interfaces/Static/IDataDragonEndpoints.cs
@@ -1,7 +1,7 @@
 namespace RiotSharp.Endpoints.Interfaces.Static
 {
     /// <summary>
-    /// The interface containing all static endpoints
+    /// The interface containing all Data Dragon endpoints
     /// </summary>
     public interface IDataDragonEndpoints
     {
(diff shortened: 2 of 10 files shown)


RiotSharp/Endpoints/Interfaces/Static/IDataDragonEndpoints.cs
```diff
@@ -3,7 +3,7 @@ namespace RiotSharp.Endpoints.Interfaces.Static
     /// <summary>
     /// The interface containing all static endpoints
```
JanOuborny: Update comment

RiotSharp.Test/RiotApiTest.cs
```diff
@@ -139,7 +139,7 @@ public void GetLeagueGrandmastersByQueueAsync_Test()
         {
             EnsureCredibility(() =>
             {
-                var leagues = Api.League.GetLeagueGrandmastersByQueueAsync(RiotApiTestBase.SummonersRegion, RiotSharp.Misc.Queue.RankedSolo5x5);
+                 var leagues = Api.League.GetLeagueGrandmastersByQueueAsync(RiotApiTestBase.SummonersRegion, RiotSharp.Misc.Queue.RankedSolo5x5);
```
JanOuborny: Random whitespace

RiotSharp/Endpoints/Interfaces/Static/IStaticChampionEndpoint.cs
```diff
@@ -26,5 +26,14 @@ public interface IStaticChampionEndpoint : IStaticEndpoint
...
         Task<ChampionListStatic> GetAllAsync(string version, Language language = Language.en_US, bool fullData = true);
+
+        /// <summary>
+        /// Get a champion by his id asynchronously.
+        /// </summary>
+        /// <param name="key">Champion id, e.g. "266 for Aatrox".</param>
+        /// <param name="version">Patch version for returned data.</param>
+        /// <param name="language">Language of the data to be retrieved.</param>
+        /// <returns>A champion.</returns>
+        Task<ChampionStatic> GetByIdAsync(int staticChampionId, string staticVersion, Language language = Language.en_US);
```
JanOuborny: You might want to remove the static prefix from `staticChampionId` and `staticVersion` to keep the argument naming consistent.

JanOuborny (commented):
Looks good 👍 I've only found some minor change suggestions.
Also, we should rename `StaticChampionEndpoint` to just `ChampionEndpoint`.
The namespace of the static endpoints should also be changed. But let's do it later in another PR when all other endpoints are refactored.

RiotSharp/Endpoints/Interfaces/Static/IDataDragonEndpoints.cs
Capm96: Updated in next commti (y)

Capm96: Updated in next commit

RiotSharp.Test/RiotApiTest.cs
Capm96: Nice catch :) removed in update

RiotSharp/Endpoints/Interfaces/Static/IStaticChampionEndpoint.cs
Capm96: Thanks, will keep an eye out for the next ones

Capm96:
> Looks good 👍 I've only found some minor change suggestions.
> Also, we should rename `StaticChampionEndpoint` to just `ChampionEndpoint`.
> The namespace of the static endpoints should also be changed. But let's do it later in another PR when all other endpoints are refactored.

Cool! And yep I noticed that within the DataDragon class, all the properties had this Static prefix which could be removed. Does this mean we should do a full rename throughout the solution as well? i.e. any properties that contain "Static", folder names, classes/interfaces like StaticEndpointBase : IStaticEndpoint, etc?

JanOuborny:
Technically every name which contains the world `Static` could be either renamed to DataDragon or the `Static`-part can be omited. `StaticEndpointBase` and `StaticEndpoint` should be renamed to `DataDragonEndpointBase` and `DataDragonEndpoint`. 
You can do this in this PR, but for the rest, I would prefer to do it separately.

JanOuborny (approved):

BenFradet (approved):

JanOuborny:
@BenFradet 
The PR might have been merged to early. I approved the existing changes without checking if everything was refactored. That is my fault.

BenFradet:
Ok then we'll need a follow-up pr

Capm96:
@JanOuborny I will create another PR that covers the name changes you mentioned, and you can let me know of any additional refactoring we need to do there 👍

PR #673, opened 2020-09-27
Renamed unit tests to Uniform naming

entvex:
Hello dear RiotSharp community.

I have renamed the unit tests. 

I would also suggest changing some of them so there is only one assert statement and maybe switching to NUnit or xUnit, but these things can be discussed in another issue 😊

### DIFF
--- a/RiotSharp.Test/CacheTest.cs
+++ b/RiotSharp.Test/CacheTest.cs
@@ -24,7 +24,7 @@ namespace RiotSharp.Test
 
         [TestMethod]
         [TestCategory("Cache")]
-        public void AddGet_TimeSpan_ShouldAddAndExpire_ReturnsTrue()
+        public void AddGet_TimeSpan_ShouldAddAndExpire_ReturnTrue()
         {
             Cache cache = new Cache();
             cache.Add(CacheTestBase.TestKey, CacheTestBase.TestValue, new TimeSpan(0, 0, 1));
@@ -36,7 +36,7 @@ namespace RiotSharp.Test
 
         [TestMethod]
         [TestCategory("Cache")]
-        public void AddGet_DateTime_ShouldAdd_ReturnsTheTestValue()
+        public void AddGet_DateTime_ShouldAdd_ReturnTheTestValue()
         {
             Cache cache = new Cache();
             cache.Add(CacheTestBase.TestKey, CacheTestBase.TestValue, DateTime.Now + new TimeSpan(0, 5, 0));
@@ -46,7 +46,7 @@ namespace RiotSharp.Test
 
         [TestMethod]
         [TestCategory("Cache")]
-        public void AddGet_DateTime_ShouldAddAndExpire_ReturnsTrue()
+        public void AddGet_DateTime_ShouldAddAndExpire_ReturnTrue()
         {
             Cache cache = new Cache();
             cache.Add(CacheTestBase.TestKey, CacheTestBase.TestValue, DateTime.Now + new TimeSpan(0, 0, 1));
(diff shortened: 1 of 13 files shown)


RiotSharp.Test/CacheTest.cs
```diff
@@ -14,7 +14,7 @@ public class CacheTest
 
         [TestMethod]
         [TestCategory("Cache")]
-        public void AddGet_TimeSpan_ShouldAddToTheCache_Test()
+        public void AddGet_TimeSpan_ShouldAddToTheCache_ReturnTrue()
```
BenFradet, line 17:
```suggestion
        public void AddGet_TimeSpan_ShouldAddToTheCache_ReturnsTrue()
```

RiotSharp.Test/CacheTest.cs
```diff
@@ -46,7 +46,7 @@ public void AddGet_DateTime_ShouldAdd_Test()
 
         [TestMethod]
         [TestCategory("Cache")]
-        public void AddGet_DateTime_ShouldAddAndExpire_Test()
+        public void AddGet_DateTime_ShouldAddAndExpire_ReturnTrue()
```
BenFradet, line 49:
```suggestion
        public void AddGet_DateTime_ShouldAddAndExpire_ReturnsTrue()
```

BenFradet (commented):
I'd suggest standardizing between `Return` and `Returns`

RiotSharp.Test/CacheTest.cs
entvex: Do you think we should change the name to AddGet_TimeSpanShouldAddToTheCache_ReturnsTrue

So we keep the MethodName_StateUnderTest_ExpectedBehavior() naming convention?

Thanks for reviewing my pull request 👍

BenFradet: yes, I think the more standardization the better :+1:

entvex:
As you said in an earlier comment returns have been changed to return in the tests 👍

RiotSharp.Test/CacheTest.cs
```diff
@@ -154,15 +156,15 @@ public void Keys_ShouldGetAllKeys_Test()
 
         [TestMethod]
         [TestCategory("Cache")]
-        public void Keys_ShouldGetEmptyIfEmptyCache_Test()
+        public void Keys_ShouldGetEmptyIfEmptyCache_ReturnsTrue()
```
BenFradet:
```suggestion
        public void Keys_ShouldGetEmptyIfEmptyCache_ReturnTrue()
```

RiotSharp.Test/RiotApiTest.cs
```diff
@@ -92,35 +92,7 @@ public void GetChampionRotationAsync_Test()
...
-        {
-            EnsureCredibility(() =>
-            {
-                // TODO: Properly implement League id test
-                return;
-                var leagues = Api.League.GetLeagueByIdAsync(RiotApiTestBase.SummonersRegion, "LEAGUE-ID-HERE");
-
-                Assert.IsTrue(leagues.Result.Queue != null);
-            });
-        }
```
BenFradet: why was this removed?

BenFradet (commented):
Looking good :+1:

RiotSharp.Test/RiotApiTest.cs
entvex: I removed two tests, because they both had TODO's with "Properly implement"

So i just assumed they where not in use or will be fixed because the tests with TODO's seem very old.

Please let me know if you would like them added back in :)

BenFradet: let's leave them in :+1:

entvex: I have added them back in and renamed them 👍

BenFradet (approved):

Issue #684, opened 2021-05-14
MATCH-V4 being deprecated on June 14th --> migrate to V5


selmaohneh:
The MATCH-V4 endpoint will be deprecated on June 14th, see here:

https://developer.riotgames.com/apis#match-v4

What's the plan for migrating to V5? How and when will this happen?

BenFradet:
Hello, the plan is: someone needs to write support for the v5 api

xXLAOKOONXx:
Heyho,  
I had the time to write support for v5 API: https://github.com/xXLaokoonXx/RiotSharp/commit/dc14f21f27e2d08109fc565e1eb885d5e5380d8a
  
But I am not sure about the dependencies inside the project.  
The TournamentRiotApi uses some of the models I modified for Match-V5: https://github.com/BenFradet/RiotSharp/blob/573b79c271c4de221eb8b7f2788a5bd1b3230de5/RiotSharp/TournamentRiotApi.cs#L225

  
I am not into the tournament apis at all, so I have little to no knowldege on whether those sections are actually supported right now or not. And it seems I am not able to do proper requests to test either. How should we proceed?

_EDIT: I did not touch the testing at all._

BenFradet:
can we have different models for now?

xXLAOKOONXx:
Alright, I will copy all the old match models into the tournament endpoint.  
Additionally I just realized the timeline model changed as well, i ll start on that as well.

selmaohneh:
how is the release process? will this be available from nuget? can i somehow help?

BenFradet:
nuget publishing is currently broken and we'd need someone to take it over. In the meantime you'll need to build from source

selmaohneh:
nuget publishing is broken? you mean automatic publishing?

how update releasing the nuget package manually for this huge change? i could create the nuget package for you and you put it on nuget.org? v5.0.0?

BenFradet:
that could work, feel free to open a PR to update the version :+1:

AoshiW:
Personally, I wouldn't release the full version of the library yet, but some alpha/beta versions may be fine.

1. RateLimitedRequester does not support header parameters (prevent incompatible API changes in the future)
2. Tuxedo talked about some new endpoint that will contain timestamps and replace the old endpoint to get a matchIDs

Changing the framework to .NET core 3.1  would be a great change.

selmaohneh:
[...]

PR #685, opened 2021-05-19
Add MATCH-V5 Support

xXLAOKOONXx:
This Pull request adds support for MATCH-V5:
- Matches by Puuid
- Match by MatchId
- Timeline by MatchId

Additionally the functionalities of the Tournament API should be unchanged. (But the Namespace for the returned Models changed)  

This Pull request does not include any changes towards testing. I just have a 'worked on my machine'.  

Furthermore there could be some improvement done in regards of some of the values. I did not create any new enum type. I might used a few too often times 'long' instead of 'int'. Documentation from Riot is not there yet, so there is some guessing involved.

Ticket #684


xXLAOKOONXx:
Information from Riot:

match-v4 is being deprecated on June 14th, 2021. Please use match-v5 as a replacement. match-v5 is available on Development API keys as of April 14th, production access will follow in the coming weeks/month.

The match and timeline responses are not 100% locked. There may be minor field changes, which will be documented in the method details. We'll make our best effort to extend the deprecation window if any changes happen after match-v5 becomes available for production API keys.

https://developer.riotgames.com/apis#match-v5

xXLAOKOONXx:
I did not validate if tournament api is still working as I can not test this!

BenFradet (approved):
Thanks a lot, that looks good to me :+1:

That PR makes me wonder if we should ditch tournament support, wdyt?

xXLAOKOONXx:
On twitter riot gave some information on theire plans: https://twitter.com/RiotGamesDevRel/status/1395091864408129536

"We're actively working on solutions to ensure continued support for tournament codes [...]"  
so there might be an update in that regards in near future.

I do not understand the current Implementation of the tournament match portion. I found a url snippet that is used but I can not relate it to any current API provided: https://github.com/BenFradet/RiotSharp/blob/573b79c271c4de221eb8b7f2788a5bd1b3230de5/RiotSharp/TournamentRiotApi.cs#L33

So it might already be the case that there is no support for getting tournament match information (getting codes etc looked fine).

My proposal is to keep the tournament api as is and once Riot adds a new requesturl for tournament matches this requesturl should be implemented in the regarding endpoint (most likely MatchEndpoint) and not costum crafted into the TournamentRiotApi class.

Another option would be to dump all tournament support. This would clearify the identity of RiotSharp from 'reflecting riot apis' towards 'making data from riot apis available', which is what most framework users are here for, i guess. For this decision I do not have an insight in whether the tournament part is actually in use by someone.

BenFradet:
>My proposal is to keep the tournament api as is and once Riot adds a new requesturl for tournament matches this requesturl should be implemented in the regarding endpoint (most likely MatchEndpoint) and not costum crafted into the TournamentRiotApi class.

I think that makes the most sense for now, thank you for the investigation :+1:

selmaohneh:
any news here? can this be merged?

BenFradet:
I was waiting on @JanOuborny 's review but we can merge now

PR #700, opened 2021-10-13
Introduce field GameEndTimestamp that was introduced in patch 11.20

Nicolas-Ding:
GameEndTimestamp was introduced in patch 11.20 (https://developer.riotgames.com/apis#match-v5/GET_getMatch)


BenFradet (approved):
Thanks :+1:

bugproof:
Read my comments https://github.com/BenFradet/RiotSharp/commit/47b70d8e4eade5e55f88a745c7e28b76575d2a90
The commit doesn't fully address 11.20 changes.

PR #703, opened 2021-11-06
Fix so tests pass

### LINKED ISSUE #611: Fix unit tests, which are using ids
Unit tests, which are using some kind of encrypted id are not working for different api keys.

Ref:

> The API uses 3 types of IDs for players: summoner IDs, account IDs, and PUUIDs. Different APIs use different ID types, and you should use whichever type is required by the API you are using. Summoner and account IDs are only unique per region, and PUUIDs are unique globally. As of API v4, all IDs are encrypted using encryption keys unique to each project. An ID obtained with your dev key will not work with your production key (and vice versa).

MingweiSamuel:
- Update old out-of-date tests.
- [Ignore] a few tests with old schemas.
- Remove use of key-specific encrypted IDs.

Fix #611

It's a bit sloppy but better than having broken tests.

### DIFF
--- a/RiotSharp.Test/CommonTestBase.cs
+++ b/RiotSharp.Test/CommonTestBase.cs
@@ -15,20 +15,15 @@ namespace RiotSharp.Test
         public static string FaultyApiKey = "deadbeef-dead-beef-dead-beefdeadbeef";
 
         public static long InvalidSummonerId = -1;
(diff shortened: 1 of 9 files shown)


BenFradet (approved):
Thanks, good effort :+1:

PR #709, opened 2022-05-10
fixed GetLeagueEntriesAsync was not passing 'page' param correctly

Ponita0:


MingweiSamuel:
Yeah we raced on this, same as #708 (but I didn't actually test if #708 compiled)

Ponita0:
me neither lol

